PR #17, opened 2019-11-07
initial checkin for basic information 

sgrg1:
just a skeleton code for this requirement. Let me know what else you want me to add.


src/App.scss
```diff
@@ -7,14 +7,14 @@
 }
 
 .App-header {
```
martypowell: I think all of the css can probably jsut be removed.

src/BasicInformation.jsx
```diff
@@ -0,0 +1,55 @@
+import React from "react";
+import { Formik, Field, Form, ErrorMessage } from "formik";
+import * as Yup from "yup";
+
+const BasicInformation = () => {
```
martypowell: I think we want to put this in teh folder structure components/sections or something along those lines. @tmccoy529 thoughts?

src/BasicInformation.jsx
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    <Formik
+      initialValues={{
```
martypowell: can we put intiailValues, validationSchema, and onSubmit into their own const above the form please?

src/BasicInformation.jsx
tmccoy529: Yeah I think this would be a component inside of the components folder.

src/App.scss
sgrg1: removed

src/BasicInformation.jsx
sgrg1: done

src/BasicInformation.jsx
sgrg1: done. let me know what you think

src/components/BasicInformation.jsx
```diff
@@ -0,0 +1,56 @@
...
+        <Field id="address" name="address" type="text" />
+        <ErrorMessage name="address" />
+      </p>
+
+      <button type="submit">Submit</button>
+    </Form>
+  );
+};
+
+export default withFormik({
```
martypowell: I was just expecting something like

```js
const validationSchema = () => {
  Yup.object()...
}

const handleSubmit = () => {
}



```

above the form. That will allow us to move the variables into their own files, if this file starts getting too big. Does that make sense?

Issue #14, opened 2019-11-07
Create Monthly Interval Selection


martypowell:
![image](https://user-images.githubusercontent.com/1933400/68410711-2f907400-0157-11ea-80d1-ae63c2d133ca.png)

**Label** - Month of Return
**Value** - Date / Year selector - required

**Additional Requirements**

Below the label and the value we will display the following:

- **Due Date** - text - will be the end of the following month. Example If you are filing for January 2019, your due date will be Feb 28th. If you are filing for Feb 2019 your due date is March 31st.

- **Past Due** - text - if the current date is greater than the due date, we want to display a message to the user of how many months they are past due. If your due date is January 19 and you are submitting the form on March 2nd 2019 you are 1 month past due.

**Other Ideas**
If the user is prior to the due date, should we let them know how many more days are left?

PR #20, opened 2019-11-08
Create Monthly Interval Selection

martypowell:
Addresses #14 

Pending the completion of #19 

- Adds a component based on the issue listed above
- Can be used monthly or quarterly
- If quarterly the first month will pick the next two months in sequence, right now editable (but we can go back to this)

This pr does not address this component interacting with other components.


martypowell:
@baltimorecounty/front-end-developers after merging with @sgrg1 code, the layout of the form is a little wild, we can resolve that after this and @tmccoy529 #21 pr's are completed

PR #21, opened 2019-11-08
Payment toggle

tmccoy529:
Added payment toggle for monthly and quarterly along with a radio button common component and created constants for the text and api values for those options.


src/App.js
```diff
@@ -1,24 +1,12 @@
...
-        <a
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
+      <PaymentOptions></PaymentOptions>
```
martypowell: this can just be `<PaymentOptions />`

the case we want it to open and close like you have it, is if there are children.

src/common/Constants.js
```diff
@@ -1,3 +1,16 @@
+export const PaymentToggle = {
```
martypowell: I think you could simplify this to 

```js
const PaymentInterval = {
  Monthly: 2,
  Quarterly: 1
};
```

Note the 2 and 1 are switched here, per the ap docs.

image

src/common/Constants.js
```diff
@@ -1,3 +1,16 @@
...
+  PaymentFirstToggleValue: 1,
+  PaymentSecondToggleValue: 2,
+  PaymentFirstToggleText: "Monthly",
+  PaymentSecondToggleText: "Quarterly"
+};
+
+export const PaymentDirections = {
+  PaymentLabel: "Do you file your return....",
```
martypowell: I have a "labels" constant in another pr, which we can resolve in the future. (Not Blocking) maybe a notes and lables object makes sense.


Also can we stick ot either export const or const and exporting at the bottom? I actually don't care that much, ive always just used the export at the bottom, so it's easy to see the list of functions being exported. Sometimes you have functions that aren't exported and you have to sift through them if the file is large.

src/common/RadioButton.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+
+export const RadioButton = ({
```
martypowell: just curious, is this pattern coming from somewhere?

src/App.js
tmccoy529: done

src/common/Constants.js
tmccoy529: done

src/common/Constants.js
tmccoy529: agreed, Ive converted these to exports at the end of the file.

src/common/RadioButton.js
tmccoy529: radio button I got from stack overflow.

https://stackoverflow.com/questions/57566468/formik-radio-values-not-freshest-from-within-a-callback

also removed export and added export to end of file.

Issue #7, opened 2019-11-07
Add Gross Occupancy Tax Collected Form Field


martypowell:
The Gross Occupancy Tax Collected section will be added below the Monthly / Quarterly toggle.

**Label** - "Gross Occupancy Tax Collected"
**Input** - type = currency

Let's try using https://github.com/jsillitoe/react-currency-input#readme for a currency input.

![image](https://user-images.githubusercontent.com/1933400/68397933-f3064d80-0141-11ea-93ad-aed52ae4ff03.png)

It might be worth creating a generic component to share the markup for any label to the left of a currency input.

As part of this work, we should account for a **monthly** or **quarterly** form.

martypowell:
@sgrg1 moving this back and taking control of this to ensure we have three inputs when quarterly is selected.

martypowell:
@baltimorecounty/front-end-developers moved back in progress to add the other sections.

PR #27, opened 2019-11-08
initial checkin

sgrg1:
basic component only . Please let me know what else you want me to add on this one


martypowell:
> basic component only . Please let me know what else you want me to add on this one

@sgrg1 please make sure these pull requests, have a title that describes the purpose of the pull request. An description that a minimum includes the issue you are addressing (if one exists). In this case it would be #7

src/components/GrossOccupancy.jsx
```diff
@@ -0,0 +1,15 @@
+import React, { Component } from "react";
+import CurrencyInput from "react-currency-input";
+class GrossOccupancy extends Component {
```
martypowell: since we will be using a similar component again. 

image

Can we name this more generically, Something like "CurrencyInput" and instead of making the label static, pass in the value as a prop?

martypowell:
@sgrg1 you want to make sure to get the latest from integration, as it changes quite a bit.

src/components/GrossOccupancy.jsx
sgrg1: done

Issue #28, opened 2019-11-12
Cleanup Form


martypowell:
- Move App.js Logic to a form component so that we can use it with routing
- Reorder form elements based on wireframe order
- Update Return Date Selector to use `PaymentInterval` Constant

PR #29, opened 2019-11-12
Cleanup form

martypowell:
Addresses #28

- Moves app.js stuff into a TransientTaxForm compoment
- Updated app.js to include routing. Default route will be the form from bullet 1. In a future update we will add the confirmation page
- Moved basic information fields into the form component (for now) so that worked with formik.
- Removed duplicate formik stuff in PaymentOptions component
- Used PaymentOptions selected option to control the display of the ReturnDateSelector


src/components/PaymentOptions.jsx
```diff
@@ -1,39 +1,25 @@
...
 
-const handleChange = changeEvent => {
-  const { name, value } = changeEvent.target;
-  console.log("input changed", name, value);
-};
-
-const PaymentOptions = () => (
-  <div>
+const PaymentOptions = props => (
+  <React.Fragment>
```
tmccoy529: Just curious why we are removing <formik> tag here.

tmccoy529: formik......

martypowell: This is creating another instance of Formik, which is in the TransientTaxForm compoment. This causes the "paymentInterval" field to be tracked inside of this "Formik" from instead of the parent. Since we want to track all the data in one form the Formik tag was removed.

Basically the form looked like this, if that nonsesne didn't help.

- Formik
  - Account Number
  - Business Name
  - Address
  - Formik
     - paymentInterval

The change adjusts this to:

- Formik
  - Account Number
  - Business Name
  - Address
  - paymentInterval

tmccoy529: okay that makes sense.

PR #33, opened 2019-11-12
Exemption component

tmccoy529:
Added the exemption component to calculate exemption totals. This is using react-currency-input to format the text values into currency. Also added a util class to parse out the special characters.


src/App.scss
```diff
@@ -32,6 +32,22 @@ $default-border: 1px solid #e0e0e0;
   }
 }
 
+.tt_exemption-inputs {
```
martypowell: this can probably be named more generically, but i don't have a name 🤷‍♂

src/common/ParseCurrency.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Remove special characters ($ and ,) so we are working with a real float and not a string.
+ * @param {number} value dollar amount to be parsed
+ */
+const ParseCurrency = value => {
```
martypowell: You can simplify this in this case.


```js
const ParseCurrency = value => parseFloat(value.replace(/[,$]/g, ""));
```

martypowell: also adding a test file for this is probably useful, in the case we find a regression like you did when it didn't remove all commas.

src/components/forms/TransientTaxForm.jsx
```diff
@@ -56,7 +57,7 @@ const TransientTaxForm = props => (
           {/* End of Basic Information Section */}
           <PaymentOptions />
           <ReturnDateSelector paymentInterval={paymentInterval} />
-          <button type="submit">Submit</button>
+          <TaxExemptions />><button type="submit">Submit</button>
```
martypowell: Looks like you got an extra carrot in here.

src/components/TaxExemptions.jsx
```diff
@@ -0,0 +1,71 @@
...
+    const currentValue = ParseCurrency(value);
+
+    if (target.id === "ExemptionTransient") {
+      setExemptionTransient(currentValue);
+    } else {
+      setExemptionOfficial(currentValue);
+    }
+  };
+
+  useEffect(() => {
```
martypowell: so I think that I overthought this.

I think 

```jsx
const [total, setTotal] = useState(0);
```

can just become 

```js
const total = exemptionTransient + exemptionOfficial;
```

This would need to go below the two state declarations. and then you can remove the useEffect.

I believe this works, because the component will rerender once the state is set.

Let me know if this doesn't make sense.

src/App.scss
tmccoy529: yeah we also have a css class tt-month-pickers that adds display:flex as the only property. That probably needs to be renamed also. We might want to go through all the css and rename accordingly.

martypowell: I agree, not blocking.

src/components/TaxExemptions.jsx
tmccoy529: hahaha yeah, that was a little obvious after you pointed that out. I removed the use effect and this is a lot more betterer now.

src/common/ParseCurrency.js
tmccoy529: done

src/components/forms/TransientTaxForm.jsx
tmccoy529: removed

src/components/TaxExemptions.jsx
```diff
@@ -0,0 +1,68 @@
...
+  ExemptionOption1,
+  ExemptionOption2,
+  ExemptionTotal
+} = Labels;
+
+const Exemptions = props => {
+  const [exemptionTransient, setExemptionTransient] = useState(0);
+  const [exemptionOfficial, setExemptionOfficial] = useState(0);
+
+  const handleExemptionChange = onChange => {
```
martypowell: Can you try this?

```js
const handleExemptionChange = (currencyValue, currencyAsNumber )=> {
```

I think this will eliminate any need for parsing the currency and simplify this a bit?

tmccoy529: done

PR #34, opened 2019-11-13
Adjust Form Layout

martypowell:
Makes changes to structure and some minor css changes to make the layout match the wireframe more closely.


src/components/ReturnDateSelector.jsx
```diff
@@ -79,7 +81,7 @@ const ReturnInterval = props => {
   }
 
   return (
-    <div className="tt_form-section">
+    <React.Fragment>
       <div className="tt_form-group flex-end">
```
tmccoy529: So why add react.fragment here but up above in GrossOccupancy you removed the react-fragment?

martypowell: I added all the sections, to ensure we could add smaller groups to a section, in this case. 

image

Two smaller components existed in this section.

React.Fragment was used in this case, to contain elements that did not require a parent element. If they did, React.Fragment would have been a div.

martypowell: @tmccoy529 let me know if that makes sense.

src/components/GrossOccupancy.jsx
```diff
@@ -1,11 +1,12 @@
...
 import { Labels } from "../common/Constants";
+import React from "react";
+
 const GrossOccupancy = props => (
-  <React.Fragment>
+  <div className="tt_form-group">
     <label htmlFor="">{Labels.CurrencyInput} </label>
     $<CurrencyInput decimalSeparator="." thousandSeparator="," />
-  </React.Fragment>
+  </div>
```
tmccoy529: This isn't yours but for consistency the $ should be added as a prefix option inside of the currency component.  <CurrencyInput
            prefix="$"

martypowell:
@tmccoy529 this is ready for another look

martypowell:
@tmccoy529 just a note, this will need to change, to use your currency component. But will be submitted after the gross occupancy stuff.

PR #35, opened 2019-11-13
Converted to a currency control for consistency

tmccoy529:
For consistency converted the currency input component into a new component with default values. We only need to pass ID value and change events to it instead of other parameters that will always stay the same.


src/components/CurrencyControl.jsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import CurrencyInput from "react-currency-input";
```
martypowell: import ReactCurrencyInput

src/components/CurrencyControl.jsx
```diff
@@ -0,0 +1,21 @@
...
+
+const CurrencyControl = props => {
```
martypowell: CurrencyInput

martypowell: we can do this, to keep things simple, people don't need to know what currency input we are using.

src/components/CurrencyControl.jsx
```diff
@@ -0,0 +1,21 @@
...
+  const { value, id, onChangeEvent } = props;
+  return (
+    <React.Fragment>
```
martypowell: Don't think you need React.Fragment here, i don't believe it's doing anything.

src/components/CurrencyControl.jsx
tmccoy529: done

src/components/CurrencyControl.jsx
tmccoy529: removed

src/components/CurrencyControl.jsx
tmccoy529: Done

PR #38, opened 2019-11-13
Add a Generic Component for Currency Input

martypowell:
Add a component that takes in a label and name and ensures the data is pushed to the form's data context. Utilizes formik's field to handle setting the values into the form.

The data will be pushed into it's object by name, and each value for that field will be captured based on it's index.

**Examples**

**Monthly**

```js
{
  0: 25.00 /* month */
}
```

**Quarterly**

```js
{
  0: 25.00, /*first month in quarter*/
  1: 50.00, /* second month in the quarter */
  2: 100.000 /* third month in the quarter */
}
```

This was setup this way in the hopes we could include the calculations regardless of how they are included.


martypowell:
@tmccoy529 there are probably some ui tweaks here required, but I was trying to get this in since it was "working"

PR #39, opened 2019-11-13
Convert Exemption to Generic Field

martypowell:
- Converted Exemption Input Fields to use Generic Component
- Added a component to handle totals when a payment interval is involved


src/components/PaymentTotal.jsx
```diff
@@ -0,0 +1,40 @@
...
+              <span className="tt_total">${total.toFixed(2)}</span>
+            </div>
+          );
+        })}
+      </div>
+    </div>
+  );
+};
+
+export default PaymentTotal;
```
martypowell: add some prop-types here.

martypowell: done

PR #41, opened 2019-11-13
Net room rental

tmccoy529:
Added New Room Rental section and cleaned up unused components and unused CSS. Also added CSS to include calendar icons in the calendar fields.


src/App.scss
```diff
@@ -27,26 +27,23 @@ $default-border: 1px solid #e0e0e0;
...
-  display: flex;
-  margin-bottom: $default-spacing;
-  justify-content: space-between;
+  label {
+    display: block;
+  }
 }
 
-.tt_exemption-totals {
+.total {
```
martypowell: I think we'll want this to be more specific, but we can clean this up later.

src/components/ReturnDateSelector.jsx
```diff
@@ -105,6 +105,9 @@ const ReturnInterval = props => {
                   dateFormat="MM/yyyy"
                   showMonthYearPicker
                 />
+                <div className="tt_calendar-icon">
```
martypowell: not sure you need this container element here. but well leave for now.

Issue #44, opened 2019-11-14
Format Currency Labels as Currency


martypowell:
Currently the PaymentTotal component does not display total's properly. An example of this is `$1111111.11` should be `$1,111,111.11`

martypowell:
@baltimorecounty/front-end-developers i'll take this one, since it's impacted by the negative issue @tmccoy529 discovered.

PR #46, opened 2019-11-14
Allow Negative PaymentField and other changes

martypowell:
Addresses issue where Exemptions were not being considered negative despite @tmccoy529 changes.

- Addresses #44 in order to display totals negatively (adds FormatCurrency utility)
- Adds `isNegativeValue` prop to PaymentField to say that field should be treated as a negative


src/components/PaymentField.js
```diff
@@ -9,7 +9,12 @@ const CustomInputComponent = ({
   ...props
 }) => {
   const { name } = field;
-  const { paymentInterval, label, buildMonthLabel = () => {} } = props;
+  const {
+    isNegativeValue,
```
martypowell: you should add prop types here.

martypowell: added proptypes for this component.

src/components/PaymentField.js
```diff
@@ -32,11 +38,15 @@ const CustomInputComponent = ({
...
         {monthsToSelect.map((month, monthIndex) => {
           const inputName = `${name}-${monthIndex}`;
+          const inputValue = values[monthIndex]
+            ? Math.abs(values[monthIndex])
+            : 0;
+
           return (
             <div className="tt_month-picker" key={inputName}>
               <label htmlFor={inputName}>{buildMonthLabel(monthIndex)}</label>
               <CurrencyInput
```
martypowell: This should use @tmccoy529 wrapper component.

martypowell: nevermind, it went bye bye

PR #47, opened 2019-11-14
Identification section from#25

sgrg1:
just a skeleton code without CSS.


src/App.js
```diff
@@ -3,12 +3,14 @@ import { HashRouter as Router, Route } from "react-router-dom";
...
 
 function App() {
   return (
     <div className="tt_app">
-      <Router>
-        <Route exact path="/" component={TransientTaxForm} />
+       <Router>
+        <Route exact path="/" component={TransientTaxForm} /> 
+        <Route exact path="/" component={IdentificationSection} />
       </Router>
```
tmccoy529: I think this is just added to the bottom of the tax form correct @martypowell. I don't think this needs its own route.

src/components/forms/IdentificationSection.jsx
```diff
@@ -0,0 +1,76 @@
...
+    onSubmit={onSubmit}
+  >
+    {props => {
+    
+      return (
+        <Form>
+          <div className="tt_form-section">
+            <label htmlFor="submittedBy">Return Submitted By</label>
+            <label htmlFor="labelQuestion">{Labels.Question}</label>
+            <label htmlFor="Date">Date:{getDate()}</label>
```
martypowell: There is a date utility `GetFormattedDueDate` that you can use, instead of creating this new getDate function.

src/components/forms/IdentificationSection.jsx
```diff
@@ -0,0 +1,76 @@
...
+  console.log(values);
+  setSubmitting(false);
+};
+const getDate = ()=>{
+  let today = new Date();
+  return  (today.getMonth() + 1) +'/' + today.getDate() +'/' + today.getFullYear();
+}
+
+const IdentificationSection = props => (
+  <Formik
```
martypowell: you shouldn't need the formik tags here, this is going to be included in the main section. If we include duplicate formik tags it's harder to manage the data.

src/components/forms/IdentificationSection.jsx
```diff
@@ -0,0 +1,76 @@
+import React from "react";
```
martypowell: it' seems like you created a whole other form, but this section is still part of the main form `TransientTaxForm`. It is ok to move it to it's own file, but it doesn't need all the other formik stuff, unless you are using it specifically.

src/common/Constants.js
```diff
@@ -16,6 +16,8 @@ const RatesAndFees = {
 };
 
 const Labels = {
+  Question:"Why do we need this?",
```
martypowell: A better property name for this would be best. Maybe `WhyDoWeNeedThis`

src/App.js
```diff
@@ -3,12 +3,14 @@ import { HashRouter as Router, Route } from "react-router-dom";
...
+import IdentificationSection from "./components/forms/IdentificationSection"
 
 function App() {
   return (
     <div className="tt_app">
-      <Router>
-        <Route exact path="/" component={TransientTaxForm} />
+       <Router>
+        <Route exact path="/" component={TransientTaxForm} /> 
+        <Route exact path="/" component={IdentificationSection} />
```
martypowell: The only other route, I suspect we'll have for now is a confirmation page, so this can be removed.

src/App.js
martypowell: @tmccoy529 correct.

src/components/forms/IdentificationSection.jsx
sgrg1: still need help on this one , because it is not giving me current date

src/components/forms/IdentificationSection.jsx
sgrg1: done

src/common/Constants.js
sgrg1: done

src/App.js
sgrg1: done

src/components/forms/IdentificationSection.jsx
```diff
@@ -1,76 +1,38 @@
...
+        </div>
+        <label htmlFor="address">Address</label>
+        <label htmlFor="labelQuestion">{Labels.WhyDoWeNeedThis}</label>
+        <div>
+          <Field id="email" name="email" type="text" />
+          <ErrorMessage name="email" />
+        </div>
+      </div>
+      <div>
+        <label htmlFor="labelLegalNote">{Labels.LegalNote}</label>
```
martypowell: this should just be a <p> without htmlFor. A label should associated with some sort of input.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

src/components/forms/IdentificationSection.jsx
```diff
@@ -1,76 +1,38 @@
...
-      .transform(value => (!value ? null : value))
-      .required("Required"),
-    title: Yup.string()
-      .transform(value => (!value ? null : value))
-      .required("Required"),
-    email:Yup.string().email('Please enter a valid email address.').required('Please enter your email address.')
-  });
+const IdentificationSection = props => {
+  return (
+    <React.Fragment>
```
martypowell: Since you have `<div className="tt_form-section">` you don't need React.Fragment.

Checkout the [docs](https://reactjs.org/docs/fragments.html) for more info.

With Fragment

```jsx
<React.Fragment>
  <Component1 />
  <Component2 />
</React.Fragment>
```

Rendered Html

```html
<p>Component1</p>
<p>Component2</p>
```

or

```jsx
<div className="section">
  <Component1 />
  <Component2 />
</div>
```
Rendered Html

```html
<div className="section">
  <p>Component1</p>
  <p>Component2</p>
</div>
```

martypowell: fyi @tmccoy529 I think maybe we talked about this the other day.

src/components/forms/IdentificationSection.jsx
```diff
@@ -1,76 +1,38 @@
...
+          <ErrorMessage name="submittedBy" />
+        </div>
+        <label htmlFor="title">Title</label>
+        <label htmlFor="labelQuestion">{Labels.WhyDoWeNeedThis}</label>
+        <div>
+          <Field id="title" name="title" type="text" />
+          <ErrorMessage name="title" />
+        </div>
+        <label htmlFor="address">Address</label>
+        <label htmlFor="labelQuestion">{Labels.WhyDoWeNeedThis}</label>
```
martypowell: let's pull out the "WhyDoWeNeedThis" labels from this pr. This will be addressed #53. The reason I'm suggesting this is that it will impact this quite a bit.

src/components/forms/IdentificationSection.jsx
```diff
@@ -1,76 +1,38 @@
...
+          <Field id="submittedBy" name="submittedBy" type="text" />
+          <ErrorMessage name="submittedBy" />
+        </div>
+        <label htmlFor="title">Title</label>
+        <label htmlFor="labelQuestion">{Labels.WhyDoWeNeedThis}</label>
+        <div>
+          <Field id="title" name="title" type="text" />
+          <ErrorMessage name="title" />
+        </div>
+        <label htmlFor="address">Address</label>
```
martypowell: Per the mockup, the label should be "Email", the htmlFor will also need to match the input whose naem and id is "email"

image

src/components/forms/IdentificationSection.jsx
```diff
@@ -1,76 +1,38 @@
...
-      .required("Required"),
-    email:Yup.string().email('Please enter a valid email address.').required('Please enter your email address.')
-  });
+const IdentificationSection = props => {
+  return (
+    <React.Fragment>
+      <div className="tt_form-section">
+        <label htmlFor="submittedBy">Return Submitted By</label>
+        <label htmlFor="labelQuestion">{Labels.WhyDoWeNeedThis}</label>
+        <label htmlFor="Date">Date:{GetFormattedDueDate(new Date())}</label>
```
martypowell: this can just be a p tag. See this other [comment](https://github.com/baltimorecounty/react-transient-tax/pull/47/files/c5adf65013fbfcba0085c5b0aba30f315a5271f6..d39fee1dc3de2d7bd7dbdf16f535c094731184a3#r346806204) if you have questions.

martypowell: you are also missing a space between `Date:` and the actual date value.

should be `Date: {GetFormattedDueDate(new Date())}`

src/common/Constants.js
```diff
@@ -23,7 +23,7 @@ const RatesAndFees = {
 };
 
 const Labels = {
-  Question:"Why do we need this?",
+  WhyDoWeNeedThis:"Why do we need this?",
```
martypowell: Despite us removing the elements that use this, we can keep this in this file.

src/components/forms/IdentificationSection.jsx
```diff
@@ -1,76 +1,38 @@
...
-          </div>
-          <button type="submit">Submit</button>
-        </Form>
-      );
-    }}
-  </Formik>
-);
-
-export default IdentificationSection;
+export default connect(IdentificationSection);
```
martypowell: we do not need to use connect inside of this file. Everything we are doing with formik in this component does not require the connect call.

src/components/forms/IdentificationSection.jsx
sgrg1: done

src/components/forms/IdentificationSection.jsx
sgrg1: done

src/components/forms/IdentificationSection.jsx
sgrg1: done

src/components/forms/IdentificationSection.jsx
sgrg1: done

src/components/forms/IdentificationSection.jsx
sgrg1: done

src/components/forms/IdentificationSection.jsx
sgrg1: done

PR #49, opened 2019-11-14
Added Remitted Stuff

martypowell:
Submitting because the changes are valid, but this isn't completely working.

- Added rest of Remitted Fields
- Fixed PaymentField and PaymentTotal to use the actual months to report to attempt to fix calculations in the future
- Only show totals if the return is being filed late

The issue in this pr is that we can't calculate totals, if they are calculated only in labels.

Please ask questions and sorry about the size.


src/components/ReturnDateSelector.jsx
```diff
@@ -80,6 +80,16 @@ const ReturnInterval = props => {
...
+  useEffect(() => {
+    setMonths({});
+  }, [paymentInterval]);
+
+  useEffect(() => {
+    const { isLate, value } = status;
+    setFieldValue("monthsLate", isLate ? value : 0);
+    setFieldValue("isReturnLate", isLate);
+  }, [status, setFieldValue]);
+
```
sgrg1: on a DatePicker component on line 119 ( i believe) , since we are not displaying selected date range we may not need "selectsStart" attribute.

martypowell: @sgrg1 not need, removed.

PR #51, opened 2019-11-14
Fix "Form" Calculations

martypowell:
Payment Total Field did not work as intended unless the calculated field was a currency input. This pr looks to address that by taking this [idea](https://github.com/jaredpalmer/formik/issues/1312#issuecomment-461407815) to heart.

- Added calculation functions to handle
  - CalculateTotalsPerMonths - returns totals, based on fields that you pass into them
  - GetCalculatedTotals - returns all totals for the form
- Adjusted Payment Total to accept an object of totals instead of the data. The object will look like this for a month entry `{ 0: 20.50 }` and like this for a quarter `{ 0: 20.50, 1: 50, 2: 75 }`. The key is the index of the month on the form 0 being the first month in the quarter.
- Updated `PaymentTotal` component to handle the data changes, and made it dumber (more visual) than actually doing anything.


src/common/Calculations.js
```diff
@@ -29,4 +29,114 @@ const CalculatePenalty = (
...
+    totalInterestAndPenalties
+  };
+};
+
+export {
+  CalculateInterest,
+  CalculatePenalty,
+  CalculateTaxCollected,
+  CalculateTotalsPerMonths,
+  GetCalculatedTotals
```
tmccoy529: Yeah I like that all the logic for these calculations has been broken out of the components and put here. It makes it easier to find any logic issues and also makes the components easier to read.

PR #57, opened 2019-11-15
Whydoweneedthis#53

sgrg1:
@martypowell , Can you please check why this does not work as modal ?


src/components/forms/Modal.jsx
```diff
@@ -0,0 +1,53 @@
...
+				<Modal show={this.state.show} onHide={this.handleClose}>
+					<Modal.Header closeButton>
+						<Modal.Title><h3>Why do I need an account?</h3></Modal.Title>
+					</Modal.Header>
+					<Modal.Body><div>
+
+						<p>In order to report an issue online, we require a one-time account creation.</p>
+						<p>This allows us to better track and follow up on issues in a timely fashion, as it ensures that we get all of the proper contact information when the issue is submitted.</p>
+						<p>At this time we cannot take anonymous requests online.</p>
+
```
tmccoy529: I know this is example text, so this is probably moot but we might want to think about making content come from the constants file.

martypowell: I'm kinda confused by this, does the bootstrap modal not just work automatically without passing in these click events?

src/App.scss
```diff
@@ -9,6 +9,11 @@ $default-border: 1px solid #e0e0e0;
     margin: 0;
   }
 }
+
+.inner-render{
```
martypowell: what is this doing?

sgrg1: removed. My bad, I was testing I believe

src/components/forms/Modal.jsx
sgrg1: I believe it does not . I did look into other modal and has the same onclick events.

martypowell: ok, can we at least take the copied code and use the existing functional pattern for our components?

```jsx
const Modal = props => {
  return <p>test</p>
}
```


[this](https://react-bootstrap.github.io/components/modal/#modals-live) shows about what you are looking for

src/components/forms/Modal.jsx
```diff
@@ -0,0 +1,53 @@
+import React, { Component } from 'react';
+import Modal from 'react-bootstrap/Modal'
+import { Button } from 'reactstrap';
+class ModalDisplay extends Component {
```
martypowell: We are going to use a couple times, so it probably make sense to figure out whats going to be different in each instance (I think title, and content) and then make the component accept those things.

Example

```jsx
<InformationModal title="My Awesome Title">
  <p>My Awesome Content</p>
</InformationModal >
```

This is just a sample, not necessarily the solution.

src/components/forms/IdentificationSection.jsx
```diff
@@ -2,12 +2,16 @@ import React from "react";
...
+import Modal from "../forms/Modal";
 
 const IdentificationSection = props => {
   return (
     <div className="tt_form-section">
       <div className="tt_date-group">
-        <label htmlFor="submittedBy">Return Submitted By</label>
+        <label htmlFor="submittedBy">
+          Return Submitted By
+          <Modal />
```
martypowell: Not blocking I'm nervous about the placement here. I don't have a good suggestion otherwise.

src/components/forms/Modal.jsx
sgrg1: done. Marty let me know what you think . is this what you wanted?

src/components/forms/IdentificationSection.jsx
sgrg1: where ever you want to place. I did it for it to display the text

src/components/forms/InformationModal.jsx
```diff
@@ -0,0 +1,23 @@
...
+import { Button } from "reactstrap";
+import ModalDisplay from "./Modal";
+
+const InformationModal = props => {
+  const [modalShow, setModalShow] = useState(false);
+  const { title, content } = props;
+
+  return (
+    <React.Fragment>
+      <Button type="button" color="link" onClick={() => setModalShow(true)}>
```
martypowell: i would just move the onClick for open and close to functions.

```js
const showModal = () => {
  setModalShow(true)
};

...
<Button type="button" color="link" onClick={showModal}>
```

Check this screenshot out from react docs to see why


image

sgrg1: done

Issue #32, opened 2019-11-12
Add Exemption Certificate Form


martypowell:
If a user enter an exemption that is greater than 0, they will be required to fill out an exemption certificate

![image](https://user-images.githubusercontent.com/1933400/68677041-4ce38a80-0529-11ea-93b3-704341035a89.png)

The user should be able to enter any number of exemptions with the following information.

- From and To Date (use datepicker)
- Exemption Type See Diagram
- The form should display in a way that it isn't too cumbersome.

Messages to include

- above - I certify that the occupancy of the facilities above have been (or will be) furnished for the exclusive use of and will be paid by or from the funds of (Please indicate):
- below - That all of the qualifications explained on this form have been met so as to exempt this occupancy from the tax imposed by County Code 11-4-401.

martypowell:
@tmccoy529 I am going to come back to this today, I just need to take some time away to get some clarity.

PR #58, opened 2019-11-15
Exemption Certificate

martypowell:
Addresses #32 

- Shows exemption form if an exemption value is present
- Add, Edit and Remove Exemptions
- Exemption List is added as a formik field

Sorry for the size of this one.


src/components/ExemptionCertificate.jsx
```diff
@@ -0,0 +1,67 @@
...
+    setExemptions(exemptions.filter(exemption => exemption.id !== exemptionId));
+  };
+
+  return (
+    <div className="tt_form-section">
+      <p>
+        I certify that the occupancy of the facilities above have been (or will
+        be) furnished for the exclusive use of and will be paid by or from the
+        funds of:
+      </p>
```
tmccoy529: Should this be a constant label?

src/components/ExemptionCertificate.jsx
```diff
@@ -0,0 +1,67 @@
...
+          exemptions={exemptions}
+          handleEditClick={editExemption}
+          handleRemoveClick={removeExemption}
+        />
+      )}
+      <p>
+        That all of the qualifications explained on this form have been met so
+        as to exempt this occupancy from the tax imposed by County Code
+        11-4-401.
+      </p>
```
tmccoy529: See above.

src/components/forms/TransientTaxForm.jsx
```diff
@@ -113,7 +125,7 @@ const TransientTaxForm = props => (
             <PaymentOptions />
             <ReturnDateSelector paymentInterval={paymentInterval} />
           </div>
-          {Object.keys(monthsToReport).length > 0 && (
+          {isPaymentIntervalSelected && (
             <React.Fragment>
```
tmccoy529: nice change here, makes it easier to follow

tmccoy529:
image

Date error

tmccoy529:
image

added exemption without selecting an option

src/components/ExemptionCertificate.jsx
martypowell: @tmccoy529 yes I think that probably makes sense.

src/components/ExemptionCertificate.jsx
martypowell: @tmccoy529 yes I think that probably makes sense.

src/components/ExemptionCertificate.jsx
martypowell: @tmccoy529 let me know if what i did makes sense.

- Created a Messages object in constants
- Created a specific section for the Exemption Cert
- Used those in the form

src/components/ExemptionCertificate.jsx
martypowell: See https://github.com/baltimorecounty/react-transient-tax/pull/58#discussion_r347361622

martypowell:
> image
> 
> added exemption without selecting an option

@tmccoy529 all validation addressed.

- Handled outside of formik for now, since it was quicker
- Displays a list of errors above the exemption selector form
- Will not allow user to submit without type, fromDate, toDate

One note, we will probably need to style this, but i didn't want to get to far in the weeds so I just added red color to the list for now.

martypowell:
@sgrg1 brought up a good point, about when a user edits an exemption but does not save cahnges. I have added an [enhancement](#60) to address this.

PR #59, opened 2019-11-15
Confirmation page content

tmccoy529:
This is kind of a skeleton for now until the API is linked up as most of the data will be coming from that. I added a date utility to format date or time based on the pattern sent to it. More or less exposing the format function inside of date-fns but without having to worry about adding months for due dates. Added a ton of constant labels that will have portions of the text populated once we get the API response linked up.


tmccoy529:
the one deep scan issue is okay for now. Its a place holder for the API call. I can comment it out if preferred.

src/App.js
```diff
@@ -3,13 +3,14 @@ import { HashRouter as Router, Route } from "react-router-dom";
 import "react-datepicker/dist/react-datepicker.css";
 import "./App.scss";
 import TransientTaxForm from "./components/forms/TransientTaxForm";
-
+import ConfirmationForm from "./components/forms/ConfirmationForm";
```
martypowell: not blocking but ConfirmationPage would make more sense at least to me.

src/common/Constants.js
```diff
@@ -47,7 +47,21 @@ const Labels = {
...
-  PenaltyInterestTotal: "Total Interest and Penalties"
+  PenaltyInterestTotal: "Total Interest and Penalties",
+  ConfirmationHeader: "Your Baltimore County Transient Occupancy Tax Return",
+  ConfirmationSubHeader: "Transient Tax Return Submitted",
+  ConfirmationBody:
+    "You have successfully completed the Baltimore County Transient Occupancy Tax Return. Your confirmation number for this return is {ID}. <br><br> Please present this number to the appropriate Budget and Finance Official when making inquiries in regards to your Transient Occupancy Tax Return.", //{ID} will be replaced with the confirmation number fromt he API
+  ConfirmationNextPayment:
+    "You have signed up for {PaymentInterval} payments; the due date for your next payment is {DueDate}", //{PaymentInterval} will be replaced with the interval chosen on the form i.e. Monthly or Quarterly, {DueDate} will be replaced with the next due date
+  ConfirmationTaxDetailsHeader: "{ID} Transient Occupancy Tax Return Details:", //{ID} will be replaced with the confirmation number from the API
+  ConfirmationGridPaymentPlan: "Your Payment Plan",
```
martypowell: I think this seciton probably needs broken up. Labels was originally labels to match up with inputs. I think maybe these can just sit in the component?

src/components/forms/ConfirmationForm.jsx
```diff
@@ -0,0 +1,155 @@
...
+
+  const dateFormat = "MMMM yyyy";
+  const timeFormat = "h:mm aaa";
+  const dateToday = GetFormatedDateTime(new Date(), dateFormat);
+  const timeToday = GetFormatedDateTime(new Date(), timeFormat);
+  const confirmBody = ConfirmationBody.replace("{ID}", confirmationId);
+  const newDueDate = GetFormattedDueDate(new Date()).toString();
+
+  /*****Updates constants labels with specific data from the API response***/
+  const gridTaxRemitted = ConfirmationGridTaxRemitted.replace(
```
martypowell: I'm confused what this is doing?

src/components/forms/ConfirmationForm.jsx
```diff
@@ -0,0 +1,155 @@
...
+        <span className="">{dateToday} </span>
+        <span className="">{timeToday}</span>
+      </i>
+      <h2>{ConfirmationSubHeader}</h2>
+      <p dangerouslySetInnerHTML={{ __html: confirmBody }}></p>
+      <p>
+        <em>{nextPayment}</em>
+      </p>
+      <h3>{taxDetailsHeader}</h3>
+      <table align="left" cellPadding="1" cellSpacing="1" id="BACO_table">
```
martypowell: I think we can probably break this table to it's own component. ANd maybe even genrated teh rows and columans based on an array of values?

src/components/forms/ConfirmationForm.jsx
```diff
@@ -0,0 +1,155 @@
...
+              </p>
+            </td>
+          </tr>
+        </tbody>
+      </table>
+    </div>
+  );
+};
+
+export default connect(ConfirmationForm);
```
martypowell: I'm assuming we are using connect here to get the values?

src/components/forms/ConfirmationForm.jsx
```diff
@@ -0,0 +1,155 @@
+import React from "react";
```
martypowell: @tmccoy529 can we grab the confirmation number from the url and use that?

let me know if you have questions, about this. You'll need to update the routing in the app.js file to accept a confirmation number as a url param.

src/components/forms/ConfirmationForm.jsx
tmccoy529: Not relevant anymore but I'll explain. Text was coming from the constants file that required the confirmation ID be added. So I put a placeholder inside of the text since the constants file wouldn't have those values (guess I could have sent it up to the constants file but didn't want to start adding logic to a file that was simply holding text). So we pull the string down, search for the placeholder text and replace it with the value we wanted to see. Now that this has been pulled out of the constants file this is no longer needed.

src/components/forms/ConfirmationForm.jsx
tmccoy529: I added connect to get the values but if we are simply return all the values in the API response this wont be needed.

src/components/forms/ConfirmationForm.jsx
martypowell: ok thanks.

src/components/forms/ConfirmationForm.jsx
tmccoy529: What is the benefit of doing that instead of just getting it from the response since we have to get all the other values from the response as well?

src/components/forms/ConfirmationForm.jsx
tmccoy529: This has been broken out to be a component now instead.

src/common/Constants.js
tmccoy529: These values were removed and placed in the component

src/components/forms/ConfirmationPage.jsx
```diff
@@ -0,0 +1,70 @@
...
+  const dateFormat = "MMMM yyyy";
+  const timeFormat = "h:mm aaa";
+  const dateToday = GetFormatedDateTime(new Date(), dateFormat);
+  const timeToday = GetFormatedDateTime(new Date(), timeFormat);
+  const newDueDate = GetFormattedDueDate(new Date()).toString();
+
+  const labels = {
+    ConfirmationHeader: "Your Baltimore County Transient Occupancy Tax Return",
+    ConfirmationSubHeader: "Transient Tax Return Submitted",
+    ConfirmationBody: `You have successfully completed the Baltimore County Transient Occupancy Tax Return. Your confirmation number for this return is ${confirmationId}. <br><br> Please present this number to the appropriate Budget and Finance Official when making inquiries in regards to your Transient Occupancy Tax Return.`,
```
martypowell: I know i'm being a pain here, but i think its ok just to include it in the markup. If you don't want to do that, i would at least recommend taking "ConfirmationBody" and moving into into the component makrup so that you dont' have to call  " dangerouslySetInnerHTML"

src/App.js
tmccoy529: Changed to page

src/components/forms/ConfirmationPage.jsx
tmccoy529: agreed, removed so that we no longer need dangerouslySetInnerHTML

PR #62, opened 2019-11-18
Api Exemption Types

martypowell:
Retrieve Exemption Types from the Api, instead of constant values. This will ensure we have the proper values when submitting to the database.

### DIFF
--- a/src/common/Constants.js
+++ b/src/common/Constants.js
@@ -5,6 +5,10 @@ const DateTypes = {
   Month: "month"
 };
 
+const FormHints = {
+  NonTransient: "Sleeping accommodations from 90 consecutive days"
+};
+
 const PaymentInterval = {
   Monthly: 2,
   Quarterly: 1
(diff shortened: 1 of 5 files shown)


src/common/Constants.js
```diff
@@ -51,34 +51,6 @@ const Labels = {
   PenaltyInterestTotal: "Total Interest and Penalties"
 };
 
-const ExemptionTypes = [
-  {
-    id: 1,
-    label: "Non-Transient",
-    hint: "Sleeping accommodations from 90 consecutive days"
```
martypowell: This needs to be added I guess as a static check.

martypowell: this was added via "FormHints" constant

src/services/ApiService.js
```diff
@@ -1,5 +1,17 @@
+import axios from "axios";
+
 let exemptionId = 0;
 
+const apiBaseUrl = "//localhost:54727/api/transientTax";
```
martypowell: we will need to address this for dev and prod, but we will do this in a later change.

PR #75, opened 2019-11-20
Link up confiration page

tmccoy529:
Added the response values for the confirmation page. This will probably still need to be cleaned up a bit.

### DIFF
--- a/src/components/forms/ConfirmationPage.jsx
+++ b/src/components/forms/ConfirmationPage.jsx
@@ -8,21 +8,17 @@ import { GetTransientTaxReturn } from "../../services/ApiService";
 import { connect } from "formik";
 
 const ConfirmationForm = props => {
-  const [confirmationId, setConfirmationId] = useState([]);
+  const [confirmationNumber] = useState(props.match.params.confirmationnumber);
   const [response, setResponse] = useState([]);
 
   useEffect(() => {
-    setConfirmationId(props.match.params.id);
-  }, [props.match.params.id]);
-
-  useEffect(() => {
-    const response = GetTransientTaxReturn().then(confirmationId);
+    const response = GetTransientTaxReturn(confirmationNumber);
     setResponse(response);
-  }, [confirmationId, response]);
+  }, [confirmationNumber, response]);
 
-  const { DateSubmitted, ReturnType } = response.data;
+  const DateSubmitted = new Date();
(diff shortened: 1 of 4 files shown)


src/components/forms/ConfirmationPage.jsx
```diff
@@ -1,61 +1,105 @@
...
+      setDateSubmitted(formattedResponse.DateSubmitted);
+      setReturnMonths(monthSubmitted.trim());
+
+      return formattedResponse;
+    };
+    GetTransientTaxReturn(confirmationNumber)
+      .then(mapResponse)
+      .then(setResponse);
+  }, [confirmationNumber, props, setResponse]);
+
```
tmccoy529: I forgot to remove props and setResponse here. Will do so in an upcoming push

src/components/forms/ConfirmationPage.jsx
```diff
@@ -1,61 +1,105 @@
...
+            new Date(MonthlyData[i].Month + "/01/ " + MonthlyData[i].Year),
+            dateFormat
+          ) + " ";
+      }
+      setTotalOccupancy(totalOccupancy);
+      setTotalExemptions(totalExemption);
+      setTotalPenalties(totalPenalty);
+      setTotalRemittedTax(totalTax);
+      setDateSubmitted(formattedResponse.DateSubmitted);
+      setReturnMonths(monthSubmitted.trim());
```
martypowell: Not blocking, but this seems super cumbersome and I'd like to walk through this, this afternoon if possible. 

What I expected was the one setState at the top then a function to map the response to a new object. then once the mapping is done, set the state. That state could then be used below.

Example code:

```jsx
const [isLoading, setIsLoading] = useState(true);
const [taxReturn, setTaxReturn] = useState({});
const { field1 = "", field2 = 0 } = taxReturn;

useEffect(() => {
  GetStuff(confirmationNumber)
    .then(map)
    .then(setTaxReturn)
    .then(() => setIsLoading(false));
}, [confirmationNumber]);

return (
  <div>
    {isLoading ? <p>Loading...</p> : <div>Tax Return Stuff Goes Here {field1}</div>}
  </div>
);
```

tmccoy529: Agreed and I started it but I wanted to give us time to get this into SE so I pushed up what I had. Will do this in another branch.

Issue #65, opened 2019-11-18
Submit Return to API


martypowell:
When a valid form is present, the user should be able to submit a form to the api (not implemented yet). The API response should return  a confirmation number. The app Redirect to the confirmation page with the confirmation page as a url param. The confirmation page will then make a call to the GetReturn endpoint and display the return data to the user.

PR #76, opened 2019-11-20
Submit Form to Api

martypowell:
Addresses #65 

- Map form data to what the server wants
- Test mapping
- Updated field names to match up with server when possible
- Submit report to api, then redirect to a confirmation page

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2083,6 +2083,15 @@
         }
       }
     },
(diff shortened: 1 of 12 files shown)


martypowell:
> Hard to test but looks good. Nothing I wouldn't expect to see

@tmccoy529 hard to test because of the api?

PR #77, opened 2019-11-20
Exemption required#67

sgrg1:
displays the validation message. But I need help how not to submit form after displaying validation message


martypowell:
> displays the validation message. But I need help how not to submit form after displaying validation message

@sgrg1 let's take a look at this this afternoon.

martypowell:
clsoing in favor of #78 @sgrg1 as we discussed in a quick meeting.

Issue #63, opened 2019-11-18
Configure API Service to Handle Multiple Environments


martypowell:
Right now our API Service just points locally, this should handle, dev and production as well. We should try to use this [package](https://www.npmjs.com/package/@baltimorecounty/javascript-utilities#config-utilities) if we can. I'm not sure if it works or not.

PR #80, opened 2019-11-21
Added Trade Alias section

tmccoy529:
Addresses #63 and #80 

Added the Trade Alias field to the identification section. This hinges on if either of the two fields inside of exemption hold a value other than 0 then the Trade Alias field displays. It is not required so there is no validation against it.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -5,8 +5,25 @@ import "./App.scss";
 import TransientTaxForm from "./components/forms/TransientTaxForm";
 import ConfirmationPage from "./components/forms/ConfirmationPage";
 import { ConstantsProvider } from "./context/ConstantsContext";
+import { setConfig } from "@baltimorecounty/javascript-utilities/config";
+
+const configValues = {
+  local: {
+    apiRoot: "//localhost:54727/api/transientTax"
+  },
+  development: {
+    apiRoot: "http://testservices.baltimorecountymd.gov/api/transientTax"
+  },
+  staging: {
+    apiRoot: "http://stagingservices.baltimorecountymd.gov/api/transientTax"
+  },
+  production: {
+    apiRoot: "http://services.baltimorecountymd.gov/api/transientTax"
+  }
+};
 
 function App() {
+  setConfig(configValues);
   return (
     <div className="tt_app">
       <ConstantsProvider>
--- a/src/components/forms/IdentificationSection.jsx
+++ b/src/components/forms/IdentificationSection.jsx
@@ -1,10 +1,28 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import { Labels } from "../../common/Constants";
 import { ErrorMessage, Field } from "formik";
 import { GetFormattedDueDate } from "../../common/DatesUtilities";
 import InformationModal from "../InformationModal";
 
-const IdentificationSection = props => {
+const IdentificationSection = ({ props }) => {
+  const [showTradeAlias, setShowTradeAlias] = useState(false);
+
+  const {
+    roomRentalCollectionFromNonTransients,
+    governmentOnBusiness
+  } = props.values;
+
+  useEffect(() => {
+    if (
+      roomRentalCollectionFromNonTransients[0] < 0 ||
+      governmentOnBusiness[0] < 0
+    ) {
+      setShowTradeAlias(true);
+    } else {
+      setShowTradeAlias(false);
+    }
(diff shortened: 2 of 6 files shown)


src/components/forms/IdentificationSection.jsx
```diff
@@ -1,10 +1,28 @@
...
+    if (
+      roomRentalCollectionFromNonTransients[0] < 0 ||
+      governmentOnBusiness[0] < 0
+    ) {
+      setShowTradeAlias(true);
+    } else {
+      setShowTradeAlias(false);
+    }
+  }, [roomRentalCollectionFromNonTransients, governmentOnBusiness]);
+
```
tmccoy529: Use Effect was used here because if these values change back to 0 then we want to remove the fields again. Im also only running this validation if these fields are changed.

src/components/forms/TransientTaxForm.jsx
```diff
@@ -131,7 +131,7 @@ const TransientTaxForm = componentProps => (
           {hasExemptions && <ExemptionCertificateField />}
           {isPaymentIntervalSelected && (
             <React.Fragment>
-              <IdentificationSection />
+              <IdentificationSection props={props} />
```
sgrg1: can't we just pass   <IdentificationSection totalExemptions={totalExemptions} /> because we already have that value instead of props?

sgrg1: i mean totalExemptions variable .

tmccoy529:
@martypowell I added the environment checking to this PR as well as its small and we both worked through it.

src/components/forms/TransientTaxForm.jsx
tmccoy529: I think that is a calculated field and isnt in the props. At least I didnt see that after you suggested this. So maybe Im missing it? If it does exist and I missed it then yes, that would be a better way to handle and then I could just check for not null.

src/App.js
```diff
@@ -5,8 +5,25 @@ import "./App.scss";
...
+  staging: {
+    apiRoot: "http://stagingservices.baltimorecountymd.gov/api/transientTax"
+  },
+  production: {
+    apiRoot: "http://services.baltimorecountymd.gov/api/transientTax"
+  }
+};
 
 function App() {
+  setConfig(configValues);
```
martypowell: I'm pretty sure this can move back up out of this function again. I made an issue iwth the packages repo to investigate this why its :poop:

src/components/forms/TransientTaxForm.jsx
martypowell: I don't tink you ever want to have a `prop` called `props`.

martypowell: more to follow.

src/components/forms/IdentificationSection.jsx
```diff
@@ -1,10 +1,28 @@
...
+  const [showTradeAlias, setShowTradeAlias] = useState(false);
+
+  const {
+    roomRentalCollectionFromNonTransients,
+    governmentOnBusiness
+  } = props.values;
+
+  useEffect(() => {
+    if (
+      roomRentalCollectionFromNonTransients[0] < 0 ||
```
martypowell: this needs to use similar logic to the exemption validation, but i need to get back to you on how this may work.

src/components/forms/TransientTaxForm.jsx
martypowell: per our chat we can use `HasAtLeast1Exemption` inside of the form, and then pass that value as a prop to the identification section.

src/components/forms/IdentificationSection.jsx
martypowell: This will ultimately be removed in favor of a prop.

src/components/forms/IdentificationSection.jsx
tmccoy529: no longer needed

src/components/forms/TransientTaxForm.jsx
tmccoy529: this is done

src/components/forms/IdentificationSection.jsx
tmccoy529: removed

src/App.js
tmccoy529: moved up

PR #81, opened 2019-11-21
removed bootstrap style#69

sgrg1:
removed bootstrap css from modal


src/bootstrap.css
```diff
@@ -0,0 +1,946 @@
...
+ * Copyright 2011-2019 Twitter, Inc.
+ * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)
+ */
+/*! normalize.css v3.0.3 | MIT License | github.com/necolas/normalize.css */
+ html {
+  font-family: sans-serif;
+  -ms-text-size-adjust: 100%;
+  -webkit-text-size-adjust: 100%;
+}
+body {
```
martypowell: if this is the way we are going to approach this, i think we only want the classes with modal in the name. Check out this file from bootstrap https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_modals.scss for a guide.

I'm guessing there are probably some visibilty classes well have to keep as well.

src/App.scss
```diff
@@ -16,7 +16,9 @@ $default-border: 1px solid #e0e0e0;
   display: flex;
   justify-content: space-between;
 }
-
+.float-right {
```
martypowell: i believe this is a bootstrap utility so it can go in the bootstrap file.

tmccoy529: I commented it out and it didn't seem to affect modal at all. Is this even needed?

sgrg1: .float-right  is the class we use with your app.css "className="tt_date-group float-right" to display date on the right corner. I was there from the beginning I guess

tmccoy529: hahaha then yeah that's kind of important

Issue #82, opened 2019-11-21
Due Date is incorrect on Confirmation Page


tmccoy529:
![image](https://user-images.githubusercontent.com/40430541/69374742-ba539180-0c74-11ea-8b1d-cfa943b3205b.png)

When submitting a return the due date is looking at todays date only and should be looking at the first date in the MonthlyData portion of the response and calculate the due date based on that value for monthly and the last value for quarterly.

PR #83, opened 2019-11-21
Added a due date check for quarterly or monthly

tmccoy529:
Corrected due date to address #82 

Originally it was taking the current date and assuming that was the date submitted. It will now look to see the type of return (monthly or quarterly) then calculate the due date based on the return type.

### DIFF
--- a/src/components/forms/ConfirmationPage.jsx
+++ b/src/components/forms/ConfirmationPage.jsx
@@ -59,24 +59,15 @@ const ConfirmationForm = props => {
         );
       }
 
-      let dueDate = "";
-      if (ReturnType === 1) {
-        dueDate = GetFormattedDueDate(
-          new Date(
-            response.MonthlyData[2].Month +
-              "/01/" +
-              response.MonthlyData[2].Year
-          )
-        );
-      } else {
-        dueDate = GetFormattedDueDate(
-          new Date(
-            response.MonthlyData[0].Month +
-              "/01/" +
-              response.MonthlyData[0].Year
-          )
-        );
-      }
+      const monthOfReturn = ReturnType === 1 ? 2 : 0;
+
+      const dueDate = GetFormattedDueDate(
+        new Date(
+          response.MonthlyData[monthOfReturn].Month +
+            "/01/" +
+            response.MonthlyData[monthOfReturn].Year
+        )
+      );
 
       formattedResponse.monthlyOccupancy = monthlyOccupancy;
       formattedResponse.monthlyExemption = monthlyExemption;
--- a/src/components/forms/ConfirmationPage.jsx
+++ b/src/components/forms/ConfirmationPage.jsx
@@ -63,9 +63,9 @@ const ConfirmationForm = props => {
 
       const dueDate = GetFormattedDueDate(
         new Date(
-          response.MonthlyData[monthOfReturn].Month +
+          MonthlyData[monthOfReturn].Month +
             "/01/" +
-            response.MonthlyData[monthOfReturn].Year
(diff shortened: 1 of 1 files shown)


src/components/forms/ConfirmationPage.jsx
```diff
@@ -58,11 +59,31 @@ const ConfirmationForm = props => {
         );
       }
 
+      let dueDate = "";
+      if (ReturnType === 1) {
+        dueDate = GetFormattedDueDate(
```
martypowell: I was on logging a couple of things that i thought of on the way home, and saw this, we have something that handles this already, https://github.com/baltimorecounty/react-transient-tax/blob/integration/src/common/DatesUtilities.js#L29 ...we may just need to adjust or extend it. This comment is really justa reminder for tomorrow.

tmccoy529: I refactored this anyway since it seemed a bit verbose. Im not sure we really need to extend the function. It just needs a date to run its calculations on. That's about as basic as it can get. But let me know if you have a better idea.

src/components/forms/ConfirmationPage.jsx
```diff
@@ -58,11 +59,22 @@ const ConfirmationForm = props => {
         );
       }
 
+      const monthOfReturn = ReturnType === 1 ? 2 : 0;
+
+      const dueDate = GetFormattedDueDate(
+        new Date(
+          response.MonthlyData[monthOfReturn].Month +
```
martypowell: you have already destructured "MonthlyData" above so you shouldn't need response here or on line 68.

src/components/forms/ConfirmationPage.jsx
martypowell: @tmccoy529 yea, this is better. Since this pr is pretty small, can we move that map function out of this component and into the api service at a minimum? 

IF you want to get wild, you could move the mapping function to this [file](https://github.com/baltimorecounty/react-transient-tax/blob/integration/src/data/TaxReturnMapper.js) and then import that into the api service.

martypowell: not blocking, but would be nice.

tmccoy529: @martypowell I'll meet you in the middle. I agree that needs to be done but this PR is addressing a bug that is kind of important and Id rather get this in incase we get this moved into SE today. Changing the mapping might offer some unforeseen complexity so why don't we approve this, make a new task for moving the mapping and then I'll jump on that task?

src/components/forms/ConfirmationPage.jsx
tmccoy529: yup, missed that. Its removed

src/components/forms/ConfirmationPage.jsx
martypowell: works for me, captured here, https://github.com/baltimorecounty/react-transient-tax/issues/85

src/components/forms/ConfirmationPage.jsx
```diff
@@ -58,11 +59,22 @@ const ConfirmationForm = props => {
         );
       }
 
+      const monthOfReturn = ReturnType === 1 ? 2 : 0;
+
+      const dueDate = GetFormattedDueDate(
+        new Date(
```
martypowell: just realized this, the probably better way to create this date would be...

```js
new Date(MonthlyData[monthOfReturn].Year, MonthlyData[monthOfReturn].Month - 1, 1);
```

but i suspect this gets you around worrying that the month is zero based so not blocking.

Issue #85, opened 2019-11-22
Move Confirmation Page Mapping to the Api Service


martypowell:
- Move the mapping function from the ConfirmationPage component to the the mappings file
- Move the mapping function  call from the ConfirmationPage component to the the ApiService

PR #87, opened 2019-11-22
Move mapping and Add monthly Columns for Quarterly

tmccoy529:
This PR address both #85 and also a redesign of the quarterly confirmation page.

### DIFF
--- a/src/data/TaxReturnMapper.js
+++ b/src/data/TaxReturnMapper.js
@@ -54,36 +54,48 @@ const MapResponseDataForTaxReturn = taxReturn => {
   var totalMonthlyRemittedTax = 0;
 
   for (var i = 0; i < MonthlyData.length; i++) {
+    const {
+      GrossRentalCollected,
+      GovernmentExemptRentalCollected,
+      NonTransientRentalCollected,
+      PenaltyRemitted,
+      Year,
+      Month
+    } = MonthlyData[i];
+
     monthlyOccupancy = monthlyOccupancy.concat(
-      FormatCurrency(MonthlyData[i].GrossRentalCollected)
+      FormatCurrency(GrossRentalCollected)
     );
-    totalMonthlyOccupancy += parseFloat(MonthlyData[i].GrossRentalCollected);
+    totalMonthlyOccupancy += GrossRentalCollected;
 
     monthlyExemption = monthlyExemption.concat(
       FormatCurrency(
-        parseFloat(MonthlyData[i].GovernmentExemptRentalCollected) +
-          parseFloat(MonthlyData[i].NonTransientRentalCollected)
+        parseFloat(GovernmentExemptRentalCollected) +
+          parseFloat(NonTransientRentalCollected)
       )
     );
     totalMonthlyExemption +=
-      parseFloat(MonthlyData[i].GovernmentExemptRentalCollected) +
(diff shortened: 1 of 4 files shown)


src/data/TaxReturnMapper.js
```diff
@@ -23,6 +28,89 @@ const GetDataForMonth = (taxReturn, monthIndex) => {
...
+
+    monthSubmitted = monthSubmitted.concat(
+      GetFormatedDateTime(
+        new Date(MonthlyData[i].Month + "/01/" + MonthlyData[i].Year),
+        dateFormat
+      ) + " "
+    );
+  }
+
+  if (ReturnType === 1) {
```
martypowell: this line is particularly concerning, in the case that the database values can change, can we check the length of the data instead?

martypowell: probably object.keys of the months or something alogn those lines?

src/data/TaxReturnMapper.js
```diff
@@ -23,6 +28,89 @@ const GetDataForMonth = (taxReturn, monthIndex) => {
...
+    monthlyOccupancy = monthlyOccupancy.concat("$" + totalMonthlyOccupancy);
+    monthlyExemption = monthlyExemption.concat("$" + totalMonthlyExemption);
+    monthlyPenalty = monthlyPenalty.concat("$" + totalMonthlyPenalty);
+    monthlyRemittedTax = monthlyRemittedTax.concat(
+      "$" + totalMonthlyRemittedTax
+    );
+    monthSubmitted = monthSubmitted.concat("Totals");
+  }
+
+  const monthOfReturn = ReturnType === 1 ? 2 : 0;
```
martypowell: same applies here for return type

martypowell: i picutre something like

```js
const isMontly = Object.keys(total).length === 1;
```

tmccoy529: done

src/data/TaxReturnMapper.js
tmccoy529: done

martypowell:
@tmccoy529 noticed a couple things

- Can we make date header's bold
- The Bottom Right Total Bold
- Add confirmation number below "Transient Tax Return Submitted"

See here: http://staging.baltimorecountymd.gov/TransientTax/confirmation-quarterly.html

I think well probalby also want to remove the confimration number in front of Transient Occupancy Tax Return Details: but i haven't gotten confirmation

src/data/TaxReturnMapper.js
```diff
@@ -23,6 +34,135 @@ const GetDataForMonth = (taxReturn, monthIndex) => {
...
+
+  var totalMonthlyOccupancy = 0;
+  var totalMonthlyExemption = 0;
+  var totalMonthlyPenalty = 0;
+  var totalMonthlyRemittedTax = 0;
+  var totalMonthlyInterest = 0;
+  var totalMonthlyTaxCollected = 0;
+  var totalMonthlyNetRoomRental = 0;
+
+  for (var i = 0; i < MonthlyData.length; i++) {
```
martypowell: per the original comment, we shoudl really clean this up. We will want to address these along with https://github.com/baltimorecounty/react-transient-tax/issues/88.

Issue #73, opened 2019-11-20
Ensure App is Handled when service is down


martypowell:
When the service is down (don't start your local service to recreate) the app crashes because it can't retrieve the lookup values. We should probably display some sore of message that the app in unavailable and to try again later.

PR #94, opened 2019-11-24
Ensure Apps fails gracefully if there is a network error.

martypowell:
Addresses #73 

- Removed context, was causing too much trouble
- Moved filing types to the from, if the network request fails so that we can redirect to an error page
- Moved pages to their own directory

These changes were made, when I tried to load the app at home, and had no service running, and it stopped me from adding integration, like I originally set out to do.


src/pages/TransientTaxForm.jsx
```diff
@@ -0,0 +1,190 @@
+import React, { useEffect, useState } from "react";
```
tmccoy529: I would think this should be named TransientTaxPage now that its been moved to the page folder.

martypowell: yep, i agree good catch.

martypowell: naming update, i left form in there, it probably doens't matter, but I did it anyway.

Issue #96, opened 2019-11-25
Interest Calculation Wrong on Confirmation Page


martypowell:
Months late is statically set here, https://github.com/baltimorecounty/react-transient-tax/blob/integration/src/data/TaxReturnMapper.js#L95. This should be based on the due date.

PR #97, opened 2019-11-25
Confirmation Calculation Not accounting for due date

martypowell:
Addresses #96 

There was a static date passed in for months late, this adds the calculation on the due date as well.  

This is only part of the fix, an api update is also required to ensure monthly data is sorted in chronological order to ensure we get the proper due date.


martypowell:
@baltimorecounty/front-end-developers #94 should be reviewed before this one.

src/data/TaxReturnMapper.js
```diff
@@ -63,6 +65,27 @@ const MapResponseDataForTaxReturn = taxReturn => {
   var totalMonthlyTaxCollected = 0;
   var totalMonthlyNetRoomRental = 0;
 
+  const isMonthly = Object.keys(MonthlyData).length === 1;
+
+  const monthOfReturn = isMonthly ? 0 : 2;
+
+  const dueDate = GetFormattedDueDate(
+    new Date(
```
martypowell: shoudl set this once.

martypowell: refactored to only add this once.

martypowell:
> Looks good

Still something to come back to, but this addresses the calculations issues.

Issue #103, opened 2019-11-25
Date Pickers not displayed properly


martypowell:
Current:

![image](https://user-images.githubusercontent.com/1933400/69552615-bec9c400-0f6c-11ea-96de-3e3929eddc83.png)

Should display like this.

![image](https://user-images.githubusercontent.com/1933400/69552656-d5701b00-0f6c-11ea-8c25-40f95feaaa57.png)

Think this may just be a site executive thing.

martypowell:
This is causing an issue where it seems like the back button is taking you forward in time.

PR #108, opened 2019-11-25
Removed setting for month year date picker

tmccoy529:
Removed option that made date pickers display moth and year which caused issues when using the back button. #103

### DIFF
--- a/src/components/ReturnDateSelector.jsx
+++ b/src/components/ReturnDateSelector.jsx
@@ -126,6 +126,7 @@ const ReturnInterval = props => {
                     onChange={date => handleDateChange(date, monthIndex)}
                     startDate={startDate}
                     dateFormat="MM/yyyy"
+                    showMonthYearPicker
                   />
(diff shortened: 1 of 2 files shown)


src/components/ReturnDateSelector.jsx
```diff
@@ -126,7 +126,6 @@ const ReturnInterval = props => {
                     onChange={date => handleDateChange(date, monthIndex)}
                     startDate={startDate}
                     dateFormat="MM/yyyy"
-                    showMonthYearPicker
```
martypowell: i actually think this is right, its the placement of the arrow icons in se. 
image

martypowell: @tmccoy529 this is my fault, the bug was horrible.

tmccoy529: yeah sorry, I was confused. I'll see if there is another setting for this.

Issue #98, opened 2019-11-25
Style Validation Messages to match Current Site's style


martypowell:
The validation styles can be found on the Sign In form on [baltcogo](https://www.baltimorecountymd.gov/iwant/report.html#/)

![image](https://user-images.githubusercontent.com/1933400/69543868-7904ff80-0f5c-11ea-90da-6fdbd79c429c.png)

PR #109, opened 2019-11-25
Update errors component

martypowell:
Addresses #98 

Adds a custom "Field" component that wraps the error message and the label into one, so we can enable styling. This component will only work for text like inputs I think.

### DIFF
--- a/src/components/Field.jsx
+++ b/src/components/Field.jsx
@@ -1,8 +1,9 @@
 import React from "react";
 import { Field as FormikField } from "formik";
+import ErrorMessage from "./ErrorMessage";
 
 const Field = props => {
-  const { name, type, id } = props;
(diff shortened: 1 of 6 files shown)


src/components/Field.jsx
```diff
@@ -0,0 +1,34 @@
...
+      }) => {
+        const cssClasses = meta.error && meta.touched ? "tt_has-errors" : "";
+        return (
+          <div className="tt_form-field">
+            {label && (
+              <div className="tt_form-field__label">
+                <label htmlFor={id} className={cssClasses}>
+                  {label}
+                </label>
+                {infoComponent}
```
martypowell: this would be a good time to style this button.

martypowell: this has been addressed

PR #112, opened 2019-11-26
Fix Interest calculation

martypowell:
Per an email

> Let’s say I’m paying now (11/25) for the month of April. That payment would have been due May 30. 

> June, July, August, September and October should all be counted in the # of months late (5 months total) – but NOT November (currently November IS being included).

> Unless the person is submitting the payment on the last day of the month, that month is NOT to be counted towards the number of months to charge the business interest for. So if I were paying on 11/30 for the month of April, then November WOULD be included in the count of # of months late (6 months total).

- Adjusted tests to match criteria
- Updated logic to pass tests

### DIFF
--- a/src/common/Caculations.test.js
+++ b/src/common/Caculations.test.js
@@ -2,13 +2,8 @@ import { CalculateInterest, CalculatePenalty } from "./Calculations";
 
 describe("Interest Calculation", () => {
   test("should return the proper calculation when passed valid data based on the default interest rate of 1", () => {
-    const actual = CalculateInterest(100, 2, 0.1);
-    expect(actual).toEqual(21);
-  });
-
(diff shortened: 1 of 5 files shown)


martypowell:
@baltimorecounty/front-end-developers this is pending a question to the business owner, but i didn't want to forget about this.

martypowell:
@baltimorecounty/front-end-developers I confirmed with the customer this morning that we do not actually know the result for the failing test, she needs to confirm with the customer before we can proceed with this.

martypowell:
> @baltimorecounty/front-end-developers I confirmed with the customer this morning that we do not actually know the result for the failing test, she needs to confirm with the customer before we can proceed with this.

This is still pending confirmation.

PR #113, opened 2019-11-26
Exemption edit#60

sgrg1:
displaying message for edited exemption when submitted/

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -52,7 +52,7 @@
     -->
     <script>
       function getConfirmation() {
-        confirm("Press a button!...what ever message ");
+      return  confirm("You have unsaved changes in the expemption form, do you wish to continue without those changes?");
       }
       </script>
   </body>
--- a/public/index.html
+++ b/public/index.html
@@ -50,10 +50,6 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
-    <script>
-      function getConfirmation() {
-      return  confirm("You have unsaved changes in the expemption form, do you wish to continue without those changes?");
-      }
-      </script>
+  
   </body>
 </html>
--- a/public/index.html
+++ b/public/index.html
@@ -50,5 +50,10 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
+    <script>
+      function getConfirmation() {
+        confirm("Press a button!...what ever message ");
+      }
+      </script>
   </body>
 </html>
--- a/src/pages/TransientTaxFormPage.jsx
+++ b/src/pages/TransientTaxFormPage.jsx
@@ -71,7 +71,6 @@ const onSubmit = (values, history) => {
   if (!values.isExemptionFormDirty) {
     isSubmit = window.getConfirmation();
   }
-  console.log('isSubmit:' + isSubmit);
    if (isSubmit) {
      SaveReturn(values).then(({ ConfirmationNumber = 0 }) => {
        history.push(`/confirmationPage/${ConfirmationNumber}`);
--- a/src/pages/TransientTaxFormPage.jsx
+++ b/src/pages/TransientTaxFormPage.jsx
@@ -70,11 +70,11 @@ const getDirtyExcemptionConfirmation = () =>
     "You have unsaved changes in the expemption form, do you wish to continue without those changes?"
   );
 const onSubmit = (values, history) => {
-  var isSubmit = true;
+  var canSubmit = true;
   if (!values.isExemptionFormDirty) {
-    isSubmit = getDirtyExcemptionConfirmation();
+    canSubmit = getDirtyExcemptionConfirmation();
   }
-  if (isSubmit) {
+  if (canSubmit) {
     SaveReturn(values).then(({ ConfirmationNumber = 0 }) => {
       history.push(`/confirmationPage/${ConfirmationNumber}`);
     });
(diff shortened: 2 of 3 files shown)


src/pages/TransientTaxFormPage.jsx
```diff
@@ -14,6 +14,8 @@ import ExemptionsSection from "../components/forms/ExemptionsSection";
 import TransientTaxSection from "../components/forms/TransientTaxSection";
 import { GetFilingTypes, SaveReturn } from "../services/ApiService";
 
+
```
martypowell: this space can be removed.

src/pages/TransientTaxFormPage.jsx
```diff
@@ -28,7 +30,8 @@ const initialValues = {
   nameOfSubmitter: "",
   titleOfSubmitter: "",
   email: "",
-  tradeAlias: ""
+  tradeAlias: "",
+  editNotClicked: true
```
martypowell: can we call this `isExemptionFormDirty` ? check out this [link](https://docs.microsoft.com/en-us/office/vba/api/access.form.dirty(property)) for a good explanation on the desire to use this naming convention.

src/pages/TransientTaxFormPage.jsx
```diff
@@ -47,6 +50,15 @@ const validationSchema = () => {
     email: Yup.string()
       .email("Please enter a valid email address.")
       .required("Please enter your email address."),
+      editNotClicked: Yup
```
martypowell: we should not need this. this should be handled inside of the submit function.

src/pages/TransientTaxFormPage.jsx
```diff
@@ -66,9 +78,10 @@ const validationSchema = () => {
 };
 
 const onSubmit = (values, history) => {
-  SaveReturn(values).then(({ ConfirmationNumber = 0 }) => {
-    history.push(`/confirmationPage/${ConfirmationNumber}`);
-  });
+
```
martypowell: like we had yesterday, this should check to see if `values.isExemptionFormDirty === true`. In that case you should use https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_confirm to display a message of something like

"You have unsaved changes in the expemption form, do you wish to continue without those changes?" If the person selects ok, then the form should be submitted, otherwise the submit action should be cancelled.

src/pages/TransientTaxFormPage.jsx
sgrg1: done

src/pages/TransientTaxFormPage.jsx
sgrg1: done

src/pages/TransientTaxFormPage.jsx
sgrg1: done

src/pages/TransientTaxFormPage.jsx
```diff
@@ -66,9 +67,16 @@ const validationSchema = () => {
 };
 
 const onSubmit = (values, history) => {
-  SaveReturn(values).then(({ ConfirmationNumber = 0 }) => {
-    history.push(`/confirmationPage/${ConfirmationNumber}`);
-  });
+  var isSubmit = true;
+  if (!values.isExemptionFormDirty) {
+    isSubmit = window.getConfirmation();
```
martypowell: you can put the getDirtyExcemptionConfirmation in there like this:

```js
const getDirtyExcemptionConfirmation = () =>
  window.confirm(
    "You have unsaved changes in the expemption form, do you wish to continue without those changes?"
  );
```

you can then change the `isSubmit` line to use this function.

public/index.html
```diff
@@ -50,5 +50,10 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
+    <script>
+      function getConfirmation() {
```
martypowell: we do not want to include this here, this decouples the functionality for the form component and the confirmation. These should be included together, otherwise if we forget to include this when setting up our app, or it gets removed, the app will crash.

src/pages/TransientTaxFormPage.jsx
```diff
@@ -66,9 +67,16 @@ const validationSchema = () => {
...
+  }
+  console.log('isSubmit:' + isSubmit);
```
martypowell: please remove the logging.

src/pages/TransientTaxFormPage.jsx
sgrg1: done

src/pages/TransientTaxFormPage.jsx
sgrg1: removed

src/pages/TransientTaxFormPage.jsx
```diff
@@ -64,11 +65,20 @@ const validationSchema = () => {
...
-
+const getDirtyExcemptionConfirmation = () =>
+  window.confirm(
+    "You have unsaved changes in the expemption form, do you wish to continue without those changes?"
+  );
 const onSubmit = (values, history) => {
-  SaveReturn(values).then(({ ConfirmationNumber = 0 }) => {
-    history.push(`/confirmationPage/${ConfirmationNumber}`);
-  });
+  var isSubmit = true;
```
martypowell: this should really be something like `canSubmit` or `shouldSubmit`. this suggests an action can be taken. the `is` usually determining if something "is" something isRed, isActive etc etc. NOt blocking.

sgrg1: done

Issue #105, opened 2019-11-25
Add IsEditing to Exemptions list when someone is editing an exemption


martypowell:
We should also hide the buttons until the record is saved again.

**Update**, can we also add a class to the row, so that we can style it, and replace the buttons with text that says "Editing..."

sgrg1:
done

PR #114, opened 2019-11-26
Is editing exeptions#105

sgrg1:
Addresses #105

removed edit button for editing exemption

### DIFF
--- a/src/components/ExemptionListItem.jsx
+++ b/src/components/ExemptionListItem.jsx
@@ -43,7 +43,7 @@ const ExemptionListItem = props => {
             Remove
           </button>
         ) : null)}
-      {isDisabled && <span className="tt_is-editing">Editing...</span>}
+      {isDisabled && <span className="tt_is-editing tt-color">Editing...</span>}
     </li>
   );
 };
--- a/src/components/ExemptionListItem.jsx
+++ b/src/components/ExemptionListItem.jsx
@@ -32,17 +32,18 @@ const ExemptionListItem = props => {
       {label} - From: {format(fromDate, DefaultDateFormat)} To:{" "}
       {format(toDate, DefaultDateFormat)}
       {handleEditClick &&
-        (isDisabled === false ? (
+        (!isDisabled ? (
           <button className={cssClasses} type="button" onClick={editItem}>
-            {isDisabled === false ? "Edit" : "Editing..."}
+            Edit
           </button>
         ) : null)}
       {handleRemoveClick &&
-        (isDisabled === false ? (
+        (!isDisabled ? (
           <button type="button" onClick={removeItem}>
             Remove
           </button>
         ) : null)}
+      {isDisabled && <span className="tt_is-editing">Editing...</span>}
     </li>
   );
 };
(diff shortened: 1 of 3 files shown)


martypowell:
> removed edit button for editing exemption

@sgrg1 what issue is this addressing?

src/components/forms/ExemptionCertificateField.jsx
```diff
@@ -14,20 +14,26 @@ const ExemptionCertificate = ({
   const { setFieldValue } = form;
   const [exemption, setExemption] = useState({});
   const [exemptions, setExemptions] = useState([]);
+  const [isEdit, setIsEdit] =useState(0);
```
martypowell: to be consistent, this should be `isSelectorFormDirty`

sgrg1: done

src/components/ExemptionListItem.jsx
```diff
@@ -19,14 +19,16 @@ const ExemptionListItem = props => {
...
+   const cssClasses="editButton";
   return (
     <li className="tt-exemption">
       {label} - From: {format(fromDate, DefaultDateFormat)} To:{" "}
       {format(toDate, DefaultDateFormat)}
-      {handleEditClick && (
-        <button type="button" onClick={editItem}>
-          Edit
+     {handleEditClick  && (
+        <button className={cssClasses} type="button" onClick={editItem} disabled={isDisabled}>
```
martypowell: you actually had this closer before, check this out https://github.com/baltimorecounty/react-transient-tax/commit/ebbcb415181c7d4bc45926c8f3453fb79964062c

What should happen

- The row buttons (edit and remove) are hidden when a row is edited
- The row buttons will be replaced with a <span> that says "Editing..."
- The row should get a class stating the record is being edited so that we can style it
- When the record is saved, the buttons should return to their original state

src/components/ExemptionListItem.jsx
```diff
@@ -19,14 +19,16 @@ const ExemptionListItem = props => {
   const removeItem = () => {
     handleRemoveClick(id);
   };
-
+   const isLoadingText =isSelectorFormDirty !==id ? 'Edit': 'Editing....';
+   const isDisabled =isSelectorFormDirty !==id ? false: true;
```
martypowell:
```js
const isDisabled =isSelectorFormDirty !==id ? false: true;
```

is the same as

```js
const isDisabled =isSelectorFormDirty ===id;
```

martypowell: you are also using this check twice, for isLoadingText and isDisabled, this should be consolidated into one variable.

sgrg1: done

src/components/forms/ExemptionCertificateField.jsx
sgrg1: done

src/components/ExemptionListItem.jsx
```diff
@@ -20,20 +25,25 @@ const ExemptionListItem = props => {
...
+            Edit
+          </button>
+        ) : null)}
+      {handleRemoveClick &&
+        (!isDisabled ? (
+          <button type="button" onClick={removeItem}>
+            Remove
+          </button>
+        ) : null)}
+      {isDisabled && <span className="tt_is-editing tt-color">Editing...</span>}
```
martypowell: i think you mean tto apply this to `<li className="tt-exemption">` ?

sgrg1: fixed

PR #120, opened 2019-11-26
Added error messaging based on error type

tmccoy529:
Added some error custom messaging. Created a utility class to determine what type of error message we are getting and then added a switch to the error page itself to take the error type and display the proper message.

### DIFF
--- a/src/common/ErrorUtility.js
+++ b/src/common/ErrorUtility.js
@@ -1,0 +1,9 @@
+const ErrorPath = error => {
+  if (error.response) {
+    return "/error/invalidconfirmation";
+  } else {
+    return "/error/noresponse";
+  }
+};
+
+export { ErrorPath };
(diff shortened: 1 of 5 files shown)


src/common/ErrorUtility.js
```diff
@@ -0,0 +1,9 @@
+const ErrorPath = error => {
```
martypowell: You can simplify this a good bit.

```js
const ErrorPath = error => `/error/${error.response ? "invalidConfirmation" : "" } `;
```

I am not sure if you can just take in the route if there is no response or not, but that would be ideal, rather than "noresposne".

martypowell: I take my second part of that comment pack. Can we make the error "network"?

tmccoy529: done

Issue #117, opened 2019-11-26
Currency input does not work in microsoft browsers.


martypowell:
If you test in edge, the currency input does not work. We need to come up with a solution for this.  A number input might be the quickest way forward. This might be a temporary solution.

https://stackoverflow.com/a/14651044/1143670

PR #121, opened 2019-11-27
Update currency control

martypowell:
Addresses #117 

@baltimorecounty/front-end-developers this is a stab at getting a cross browser currency component. I tried to "add" commas to the input and things got a bit messy, so for now i have left this to be a number input so that everything calculates okay.

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -199,3 +199,5 @@ label.tt_has-errors {
 .react-datepicker__navigation {
   width: 0 !important;
 }
+
+@import "./styles/components/currency-input";
--- a/src/components/CurrencyInput.jsx
+++ b/src/components/CurrencyInput.jsx
@@ -1,0 +1,13 @@
+import React from "react";
+
+const CurrencyInput = props => {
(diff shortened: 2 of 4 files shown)


tmccoy529:
@martypowell I did find some issues with the input functionality in Edge and Chrome. It appears that the 0 placeholder doesn't delete when the box is selected. The reason why this is a problem is that it either adds a zero to the end of a number or to the front. You cannot delete this 0 until at least one other number is present. So you end up with either 012345.99 or worse 123450.99 then have to physically highlight the 0 and delete it. I would suggest removing the 0 placeholder and leaving it blank.

sgrg1:
currency input allows you to enter number after default 0 , which does not look right .

martypowell:
@tmccoy529 do you want to make the fix you suggested to get this one through?

tmccoy529:
@martypowell @sgrg1 I have pushed up a commit that removes the 0 placeholder which was adding some confusion. We may end up having to add a regex to this to force a currency appearance but I think for now we are good on this component.

PR #130, opened 2019-12-05
Dynamic steps

martypowell:
Implemented the concept of a step, which will represent a form panel (I think maybe `FormPanel` makes more sense?). This allows us to dynamically and and remove steps to the form. This is required for Payment when the date interval is quarterly and when we need to show the exemption form.

- Added StepList and Steps
- Converted Payment Field to use Numbers instead of Objects
- FIxed array utility to be more flexible

**Note**
- Exemptions do not currently work as designed
- Form submission will fail, as this changed the formik values format

These will both be addressed in upcoming prs


src/components/forms/MonthlyPayment-Form3.jsx
```diff
@@ -59,34 +77,36 @@ const MonthlyPaymentForm3 = props => {
...
-          governmentOnBusiness
-        ]);
-        const isPaymentIntervalSelected =
-          Object.keys(monthsToReport).length > 0;
-        const hasExemptions = Object.keys(totalExemptions).some(
-          exemptionIndex =>
-            totalExemptions[exemptionIndex] <
-            0 /** exemption totals are negative values */
-        );
+        // const {
```
martypowell: this stuff needs to be re-enabled

PR #134, opened 2019-12-09
Add autocomplete

tmccoy529:
Added the autocomplete control for addresses. Now users will search for an address and it will display their selection broken up into address parts listed out in static labels.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -10,16 +10,22 @@ const { setConfig } = Config;
 
 const configValues = {
   local: {
-    apiRoot: "//localhost:54727/api/transientTax"
+    apiRoot: "//localhost:54727/api/transientTax",
+    gisApiRoot: "http://localhost:54727/api/gis/addressLookup/"
   },
   development: {
-    apiRoot: "http://testservices.baltimorecountymd.gov/api/transientTax"
+    apiRoot: "http://testservices.baltimorecountymd.gov/api/transientTax",
+    gisApiRoot:
+      "https://testservices.baltimorecountymd.gov/api/gis/addressLookup/"
   },
   staging: {
-    apiRoot: "http://testservices.baltimorecountymd.gov/api/transientTax"
+    apiRoot: "http://testservices.baltimorecountymd.gov/api/transientTax",
+    gisApiRoot:
+      "https://testservices.baltimorecountymd.gov/api/gis/addressLookup/"
   },
   production: {
-    apiRoot: "http://services.baltimorecountymd.gov/api/transientTax"
+    apiRoot: "http://services.baltimorecountymd.gov/api/transientTax",
+    gisApiRoot: "https://services.baltimorecountymd.gov/api/gis/addressLookup/"
   }
 };
 
--- a/src/App.scss
+++ b/src/App.scss
@@ -211,3 +211,28 @@ label.tt_has-errors {
     background: yellow;
   }
 }
+
+.is-highlighted {
+  background: #e0e0e0;
+}
+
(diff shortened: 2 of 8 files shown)


src/App.scss
```diff
@@ -211,3 +211,28 @@ label.tt_has-errors {
...
+  padding: 5px 10px;
+}
+
+.autocomplete-results {
+  background: white;
+  box-shadow: 0 2px 12px rgba(0, 0, 0, 0.1);
+  opacity: 1 !important;
+  position: fixed;
+  z-index: 99999;
+}
```
tmccoy529: All of this CSS was brought over from Citysourced so it might not even be needed.

src/components/AutoCompleteField.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+  return (
+    <Autocomplete
+      name="location"
+      label="Address"
+      getItemValue={item => item.label}
+      id="location-autocomplete-input"
+      items={items}
+      placeholder="123 Amazing St"
+      renderInput={props => <input className="autocomplete-input" {...props} />}
```
tmccoy529: placeholder can be removed (as I don't think it even works)

src/App.js
```diff
@@ -10,16 +10,22 @@ const { setConfig } = Config;
...
   },
   development: {
-    apiRoot: "http://testservices.baltimorecountymd.gov/api/transientTax"
+    apiRoot: "http://testservices.baltimorecountymd.gov/api/transientTax",
+    gisApiRoot:
+      "https://testservices.baltimorecountymd.gov/api/gis/addressLookup/"
   },
   staging: {
-    apiRoot: "http://testservices.baltimorecountymd.gov/api/transientTax"
+    apiRoot: "http://testservices.baltimorecountymd.gov/api/transientTax",
```
martypowell: the http protocol is different on these addresses, is that intentional? I tend to think no, since they are the same service?

martypowell: this is not blocking, and I'm not sure its the right or not, but it might be worth setting a variable for the repeated service root.

Maybe something like 

```js
const localApiRoot = '//localhost:54727/api';
const testApiRoot = '//testservices.baltimorecountymd.gov/api';
```

src/components/AutoCompleteField.jsx
martypowell: @tmccoy529 this is a direct copy from baltcogo right?

martypowell: @tmccoy529 
image

The cursor seems to have quiet a bit of padding or margin?

src/components/forms/BasicInformationSection.jsx
```diff
@@ -1,16 +1,88 @@
...
+  const [Address, setItems] = useState([]);
+
+  const handleAddressChange = e => {
+    const { value } = e.target;
+    GetAddresses(value).then(response => {
+      setItems(response);
+    });
+  };
+
+  const handleAddressSelect = val => {
```
martypowell: It seems like any of the address logic inside of here, would make more sense in the `AutoCompleteField.jsx` if we changed the name to `AddressLookupField`. This component would be responsible for working like any of the other fields we have. An example is https://github.com/baltimorecounty/react-transient-tax/blob/integration/src/components/PaymentField.js

Since this is working, I'll defer to what you want to do, just let me know before we approve this pr @tmccoy529

martypowell:
> per our chat @tmccoy529 we are going to close this for now, to attempt to integrate this feature into the PaginateTaxForms branch.

closing

Issue #123, opened 2019-11-27
Add Address to Confirmation Page and Email Confirmation


martypowell:
I'd imagine this would go at the bottom of both the email and the confirmation page, and will likely not be centered like it is in the screenshot below.

![image](https://user-images.githubusercontent.com/1933400/69728248-252c1f00-10f2-11ea-8ac5-777f3033d261.png)

Note: The email change will need to happen inside of the service, and outside of this repo.

PR #135, opened 2019-12-09
Add address#123

sgrg1:
Addresses #123 

Add Address to Confirmation Page. I am not seeing confirmation page on my local mechine so I am assuming the text is displaying.

### DIFF
--- a/src/common/Constants.js
+++ b/src/common/Constants.js
@@ -14,6 +14,12 @@ const PaymentDirections = {
   PaymentNote:
     "Note: Must select the interval in which you have selected previously unless it's a new year."
 };
+const CountyAddress = {
+  Agency: "BALTIMORE COUNTY MARYLAND",
+  Department: "Office of Budget and Finance",
+  Street: "400 Washington Avenue, Room 150",
+  City: "Towson, Maryland 21204-4665"
+};
 
 const RatesAndFees = {
   PenaltyRate: 0.1, // 10%
@@ -65,5 +71,6 @@ export {
   Labels,
   Messages,
   PaymentDirections,
-  RatesAndFees
+  RatesAndFees,
+  CountyAddress
 };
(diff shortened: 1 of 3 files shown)


src/pages/ConfirmationPage.jsx
```diff
@@ -71,7 +72,12 @@ const ConfirmationForm = props => {
             DueDate={formattedDueDate}
             ReturnType={ReturnTypeDescription}
           />
+          <p>{Agency}</p>
```
martypowell: can you add a div around these items? so that we can style them in the future? It might even be worth creating a small component for this.

src/pages/ConfirmationPage.jsx
```diff
@@ -2,7 +2,8 @@ import React, { useEffect, useState } from "react";
 import ConfirmationTable from "../components/ConfirmationTable";
 import { GetTransientTaxReturn } from "../services/ApiService";
 import { ErrorPath } from "../common/ErrorUtility";
-
+import { CountyAddress } from "../common/Constants";
+const { Agency, Department,Street,City } = CountyAddress;
```
martypowell: I belive you can do this, to simplify

```js
import { CountyAddress: { Agency, Department, Street, City }  } from "../common/Constants";
```

src/pages/ConfirmationPage.jsx
martypowell: class name for this that makes sense (probably) is `tt_address`

src/common/Constants.js
```diff
@@ -14,6 +14,12 @@ const PaymentDirections = {
   PaymentNote:
     "Note: Must select the interval in which you have selected previously unless it's a new year."
 };
+const CountyAddress = {
```
martypowell: This name is ambiguous,  it is the BudgetAndFinanceOfficeAddress

src/pages/ConfirmationPage.jsx
martypowell: this actually doesn't work.

martypowell:

PR #137, opened 2019-12-09
Integerate auto complete

tmccoy529:
Autocomplete integrated with PaginateTaxForms. This is now its own component and not coupled with the basic info form anymore. Also modified the config file to have root variables for endpoints.


src/App.js
```diff
@@ -28,8 +28,7 @@ function App() {
   return (
     <div className="tt_app">
       <Router>
-        {/* TODO: this needs to be integrated into the multi-page app */}
-        <Route exact path="/" component={FormPage} />
+        <Route exact path="/" component={TransientTaxFormPage} />
         <Route
           exact
           path="/confirmation/:confirmationNumber"
```
tmccoy529: This page will change and most likely drop from this PR once @martypowell gets his PR in and I merge with it.

src/components/AddressLookupField.js
```diff
@@ -0,0 +1,114 @@
...
+    return (
+      _.startCase(_.camelCase(address)) +
+      `, ` +
+      _.startCase(_.camelCase(city)) +
+      `, ` +
+      _.startCase(_.camelCase(zip))
+    );
+  };
+
+  const displayErrors = classes => {
```
martypowell: not blcoking but this can be converted to 

```js
const displayerrors = classes => touched[field.name] && errors[field.name]
      ? "tt_form-field tt_has-errors"
      : classes;
```

src/components/AddressLookupField.js
```diff
@@ -0,0 +1,114 @@
...
+    });
+    setFieldValue("businessAddress", value);
+  };
+
+  const handleAddressSelect = val => {
+    setFieldValue("businessAddress", val);
+  };
+
+  const UpperCaseFirstLetter = (address, city, zip) => {
+    return (
```
martypowell: I know this existed, but can we attempt to do this without lodash. Also, zip doesn't required casing, as it should be a number.

src/components/AddressLookupField.js
```diff
@@ -0,0 +1,114 @@
...
+  const { label } = props;
+  const { setFieldValue, touched, errors } = form;
+  const [Address, setItems] = useState([]);
+
+  const handleAddressChange = changeEvent => {
+    const { value } = changeEvent.target;
+    GetAddresses(value).then(response => {
+      setItems(response);
+    });
+    setFieldValue("businessAddress", value);
```
martypowell: Just so we are clear, we are setting two formik fields, one for the address that is a string and the other is that address broken into parts?

src/components/AddressLookupField.js
```diff
@@ -0,0 +1,114 @@
...
+  const { name, value } = field;
+  const { label } = props;
+  const { setFieldValue, touched, errors } = form;
+  const [Address, setItems] = useState([]);
+
+  const handleAddressChange = changeEvent => {
+    const { value } = changeEvent.target;
+    GetAddresses(value).then(response => {
+      setItems(response);
+    });
```
martypowell: can we add error handling to this? For now, you can just log the error to the console, using `console.error`

src/components/AddressLookupField.js
```diff
@@ -0,0 +1,114 @@
...
+    );
+  };
+
+  const displayErrors = classes => {
+    return touched[field.name] && errors[field.name]
+      ? "tt_form-field tt_has-errors"
+      : classes;
+  };
+  const items = Address.map((item, index) => ({
+    id: item.Latitude + item.Longitude,
```
martypowell: This will need to be updated, but I need to update the service, let me do that real quick.

src/components/AddressLookupField.js
```diff
@@ -0,0 +1,114 @@
...
+    label: UpperCaseFirstLetter(item.StreetAddress, item.City, item.Zip),
+    street: item.StreetAddress,
+    city: item.City,
+    zip: item.Zip
+  }));
+
+  return (
+    <div>
+      <div className="tt_form-field__label">
+        <label htmlFor={name} id="addressSearch" className={displayErrors()}>
```
martypowell: did you mean to pass something to displayErrors here?

src/components/AddressLookupField.js
martypowell: This will end up being item.AddressId, will submit the change soon.

src/components/AddressLookupField.js
tmccoy529: Added error handling and also a null check that was tossing an error when using the backspace.

src/components/AddressLookupField.js
tmccoy529: This has been updated with AddressId

src/components/AddressLookupField.js
```diff
@@ -0,0 +1,127 @@
...
+        });
+    }
+    setFieldValue("businessAddress", value);
+  };
+
+  const handleAddressSelect = val => {
+    setFieldValue("businessAddress", val);
+  };
+
+  const UpperCaseFirstLetter = value => {
```
martypowell: I think maybe this can be simplified?

See https://stackoverflow.com/a/49366757/1143670

src/components/AddressLookupField.js
```diff
@@ -0,0 +1,127 @@
...
+  };
+
+  const toggleErrorClasses = classes => {
+    return touched[field.name] && errors[field.name]
+      ? "tt_form-field tt_has-errors"
+      : classes;
+  };
+  const items = Address.map((item, index) => ({
+    id: item.AddressId,
+    label: `${UpperCaseFirstLetter(item.StreetAddress)}${UpperCaseFirstLetter(
```
martypowell: I think you probably want to handle the spaces inside of the label, rather than the uppercasefirstletter function

tmccoy529: item.streetaddress sends a string that looks like this (400 maple leaf way). If handled in the label, it would return as 400MapleLeafWay Towson 12345. Adding the space in the function handles multiple named streets and the street number. Otherwise we get this mess.

src/components/AddressLookupField.js
tmccoy529: That wont work because that is only checking for the very first char of the string. In this case that will most likely be a number and this also wont handle street names with multiple names. Like 400 maple leaf way would be 400 maple leaf way with this solution. We want every portion of the address to start with a capital (or do we care about any of this formatting?)

src/components/AddressLookupField.js
tmccoy529: The label has no class to start so I could pass null or "" if need be.

src/components/AddressLookupField.js
tmccoy529: done

src/components/AddressLookupField.js
tmccoy529: Converted to a function

src/components/AddressLookupField.js
tmccoy529: done

src/components/AddressLookupField.js
tmccoy529: This was not true. We were only storing the label value. I have added businessAddressParts as a formik field and Ive added the selected object to that field.

martypowell: @tmccoy529 tahnks for the update, i made a small change so we could push this through to address my concern about the spaces, it's available [here](https://github.com/baltimorecounty/react-transient-tax/commit/c41ae624c17dbb4395957235062d860279428127), let me know what you think.

src/components/AddressLookupField.js
martypowell: @tmccoy529 tahnks for the update, i made a small change so we could push this through to address my concern about the spaces, it's available [here](https://github.com/baltimorecounty/react-transient-tax/commit/c41ae624c17dbb4395957235062d860279428127), let me know what you think.

martypowell: The benefit of this approach is you don't have to worry about the spaces. The function only does one thing, and that takes any word it finds, and capitalizes the first letter in each word, otherwise it's the same, no side effects.

PR #138, opened 2019-12-09
Integrate Routing into Paged Form

martypowell:
Routing was removed as part of the pages updates, this restores the original routing. This does not address specific "step" or "panel" routing.


src/App.js
```diff
@@ -24,21 +25,25 @@ const configValues = {
 };
```
tmccoy529: @martypowell I think you are missing the config values here from index.js. If you want to use the new set with the gis values I've listed it below:

const localApiRoot = "//localhost:54727/api/";
const testApiRoot = "http://testservices.baltimorecountymd.gov/api/";
const prodApiRoot = "https://services.baltimorecountymd.gov/api/";

const configValues = {
  local: {
    apiRoot: `${localApiRoot}transientTax`,
    gisApiRoot: `${localApiRoot}gis/addressLookup`
  },
  development: {
    apiRoot: `${testApiRoot}transientTax`,
    gisApiRoot: `${testApiRoot}gis/addressLookup`
  },
  staging: {
    apiRoot: `${testApiRoot}transientTax`,
    gisApiRoot: `${testApiRoot}gis/addressLookup`
  },
  production: {
    apiRoot: `${prodApiRoot}transientTax`,
    gisApiRoot: `${prodApiRoot}gis/addressLookup`
  }
};

tmccoy529: You can probably disregard this. Looking at changes it shows the config section as blank but opening the file its not. So once I add my stuff this will all update anyway.

PR #141, opened 2019-12-11
Added field set wrapper and fixed tab highlights

tmccoy529:
Added a fieldset wrapper to step.jsx. Added test css to the project to mimic the citysourced projects style, look and feel. In the process of doing that found a bug where tabs were not highlighting properly and this has been resolved in this PR as well. The CSS could probably be removed since it was there for testing of tabs but if anyone wants to test as well I left it all in.


src/App.js
```diff
@@ -3,6 +3,8 @@ import React from "react";
 import "react-datepicker/dist/react-datepicker.css";
 import { HashRouter as Router, Route } from "react-router-dom";
 import "./App.scss";
+import "./styles/SiteExecutive/citysourced-reporter.min.css";
```
martypowell: can you move these references to the `index.html` file? That way they aren't bundled in the app, and otherwise optional.

tmccoy529: done

PR #144, opened 2019-12-11
Add Review Panel

martypowell:
Updated the review step, to include the "Return Summary" which is displayed on the confirmation page.

- This required updates to the tax mapper functions, so that we could show the return summary, whether it's form data or data returned by the api.
- Also removes submit button once the form is submitting, so the user can't smash the button again. This was easier to remove than disable based on the design, and might be something to come back to.


src/components/ReturnSummary.jsx
```diff
@@ -0,0 +1,96 @@
...
+  /** Headline for the summary */
+  header: PropTypes.string,
+  /**  A list of key / value pairs with summaries data by month */
+  values: PropTypes.array,
+  /** Due date of the return */
+  dueDate: PropTypes.string,
+  /** String return type for the return (Monthly or Quarterly) */
+  returnType: PropTypes.string,
+  /** Date of return submission */
+  dateSubmitted: PropTypes.string
```
martypowell: missing paymentInterval prop

src/data/TaxReturnMapper.js
```diff
@@ -1,144 +1,209 @@
...
+  GetFormatedDateTime,
+  GetFormattedDueDate
+} from "../common/DatesUtilities";
+import { FormatCurrency } from "../common/FormatUtilities";
+
+const dateFormat = "MMMM yyyy";
+
+const sumReducer = (accumulator, currentValue) => accumulator + currentValue;
+
+const getValue = (monthlyData = [], keys = []) => {
```
martypowell: describe the methods, a more meaningful name is would also be useful.

martypowell:
@tmccoy529 good spot on the exemptions, the validation wasn't copied voer from the orginal and it was checking email isntead, which never existed in that form, so the form would never be valid.

src/data/TaxReturnMapper.js
```diff
@@ -1,144 +1,225 @@
...
+}) => {
+  const dateSubmitted = DateSubmitted
+    ? GetFormatedDateTime(new Date(DateSubmitted), "MMMM dd yyyy")
+    : "";
+  const isMonthly = Object.keys(monthlyData).length === 1;
+  const monthOfReturn = isMonthly ? 0 : 2;
+  const dueDate = monthlyData.length
+    ? new Date(
+        monthlyData[monthOfReturn].month +
+          "/01/" +
```
tmccoy529: image

Select quarterly and fill in the first two months and click for the third month to appear and this error is thrown.

martypowell: @tmccoy529 @sgrg1 good catch, this is ready for another look, I ended up setting the monthlyData object once dates are selected, then when we populate values those objects are updated. i also fixed a validation issue that (i think ) i created with this.

martypowell: @tmccoy529 apparently i submitted this comment weird and never sent yesterday at the end fo the day. @baltimorecounty/front-end-developers this is ready for another look.

tmccoy529:
image

So this happens only if you add an exemption on the first month

martypowell:
> On a Review Page , it does not display any exemption certification option you selected for the given exemption you entered. is it what it suppose to do ?

@sgrg1 I am not sure for the sake of time, we decided to leave it out.

src/components/ReturnDateSelector.jsx
```diff
@@ -97,9 +108,10 @@ const ReturnInterval = props => {
 
   useEffect(() => {
     const { isLate, value } = status;
+    setFieldValue("dueDate", dueDate);
     setFieldValue("monthsLate", isLate ? value : 0);
     setFieldValue("isReturnLate", isLate);
-  }, [status, setFieldValue]);
+  }, [status, dueDate, setFieldValue]);
 
   if (!paymentInterval) {
```
sgrg1: You may not need this if statement . Formik is validation is is checking for it on button click event. Please correct me if i am not understading.

PR #146, opened 2019-12-12
Convert App to Use Paged Forms

martypowell:
All of the changes that were required to change this to a paged form.


src/common/BuildSteps.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+const BuildSteps = (steps, activeStep) => {
```
martypowell: @tmccoy529 I think this is probably a candidate for cleanup? This is the progress bar I think?

src/components/AddressLookupField.js
```diff
@@ -0,0 +1,131 @@
...
+    setFieldValue("businessAddress", value);
+    setFieldValue("businessAddressParts", item);
+  };
+
+  /**
+   * Convert the first letter in a list of words separated by a give separator
+   * @param {string} value string that contains spaces, Ex. washington ave
+   * @param {string} separator string value for separator, Ex " ", ","
+   */
+  const UpperCaseFirstLetter = (value, separator = " ") =>
```
martypowell: I think this probably is a good candidate for moving into a utility, probably in the formatting class.

src/components/AddressLookupField.js
```diff
@@ -0,0 +1,131 @@
...
+          addressPieces.charAt(0).toUpperCase() + addressPieces.slice(1)
+      )
+      .join(" ");
+
+  const toggleErrorClasses = classes =>
+    touched[field.name] && errors[field.name]
+      ? "tt_form-field tt_has-errors"
+      : classes;
+
+  const items = Address.map((item, index) => ({
```
martypowell: this might be a good candidate to move as well, not sure if this is a mapper or jsut a formatter or both.

src/components/PaymentField.js
```diff
@@ -13,54 +13,37 @@ const CustomInputComponent = ({
...
+            prefix={`${isNegativeValue ? "-" : ""}$`}
+            decimalSeparator="."
+            thousandSeparator=","
+            allowNegative={true}
+            id={name}
+            name={name}
+            onChange={handleChange}
+            value={value}
+          />
+          {touched[field.name] && errors[field.name] && (
```
martypowell: this should be using the `ErrorMessage` component.

src/components/ReturnSummary.jsx
```diff
@@ -0,0 +1,98 @@
...
+        <tbody>
+          {values.map((item, itemIndex) => (
+            <tr key={item.id}>
+              <td colSpan="1" rowSpan="1">
+                <p>
+                  <strong>{item.key}</strong>
+                </p>
+              </td>
+              {item.value.map((result, resultIndex) => {
+                const isFirstRow = itemIndex === 0;
```
martypowell: a candidate for a refactor, mapped values could be converted to a component.

src/components/Step.jsx
```diff
@@ -0,0 +1,82 @@
+import React from "react";
+import { connect } from "formik";
+
+const Step = props => {
```
martypowell: I think this naming convention needs to change in the future to FormPanel.

martypowell:
@baltimorecounty/front-end-developers this is ready to review now.

src/components/AddressLookupField.js
martypowell: although, we may need to bundle this logic, if we move this component to some sort of library.

src/components/PaymentField.js
```diff
@@ -1,8 +1,8 @@
-import React, { useEffect, useState } from "react";
+import React, { useState } from "react";
 import PropTypes from "prop-types";
 import classnames from "classnames";
-import { Field } from "formik";
-import CurrencyInput from "./CurrencyInput";
+import { Field, connect } from "formik";
+import CurrencyInput from "react-currency-input";
```
martypowell: @tmccoy529 do you know if this is right?

I think this is actually backwards? meaning we should be using CurrencyInput instead of react-currency input.

src/components/TransientTaxTabs.jsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
```
martypowell: Added issue to make this more generic, #148

src/components/forms/BasicInformationForm.jsx
```diff
@@ -0,0 +1,64 @@
+import React from "react";
+import { Formik, Form } from "formik";
+import TransientTaxTabs from "../TransientTaxTabs";
+import Field from "../Field";
+import AddressLookupField from "../AddressLookupField";
+
+import * as Yup from "yup";
+
+const BasicInformationForm1 = props => {
```
martypowell: missed this naming here for components to match files. will in another branch and submit.

src/components/forms/MultiPageForm.jsx
```diff
@@ -0,0 +1,59 @@
...
+  const handleNavClick = stepNumber => {
+    setActiveStep(stepNumber);
+  };
+
+  return (
+    <div className="tt_form">
+      <Formik
+        initialValues={{}}
+        onSubmit={(values, actions) => {
+          setTimeout(() => {
```
martypowell: looks like this really isn't doing anything...we might have come back to this, maybe the review panel was implemented in the wrong way.

src/components/AddressLookupField.js
tmccoy529: @martypowell yeah I figured since this was more for this actual component that moving into a utility would be problematic if we decide to package it.

src/components/PaymentField.js
tmccoy529: @martypowell react-currency-input does not work in IE or Edge. So yes this needs to either be used only in Chrome and the other in Edge (not recommended from a consistency standpoint) or we need to remove it and convert to the currency input we custom made.

src/components/forms/ReviewFormPanel.jsx
```diff
@@ -0,0 +1,62 @@
...
+        <p>
+          Submitting your return. You will be redirected to a confirmation page
+          shortly...
+        </p>
+      )}
+    </form>
+  );
+};
+
+export default ReviewPanel;
```
martypowell: not blocking, i think maybe this needed to be "connect"ed to the main formik from and then submit from there. Since htis works, not recommending any action at this time.

src/pages/FormPage.jsx
```diff
@@ -0,0 +1,7 @@
+import React from "react";
```
martypowell: This looks like it's not doing anything.

src/services/ApiService.test.js
```diff
@@ -22,9 +22,9 @@ const buildReturn = (monthlyData = [], exemptions = []) => ({
   titleOfSubmitter: "Owner",
   tradeAlias: "",
   monthsToReport: { 0: new Date(2019, 0, 1) },
-  grossOccupancy: { 0: 10.0 },
-  roomRentalCollectionFromNonTransients: { 0: 20 },
-  governmentOnBusiness: { 0: 40 }
+  grossRentalCollected: { 0: 10.0 },
```
martypowell: this doesn't look valid to me anymore.

src/steps/Step.js
```diff
@@ -0,0 +1,3 @@
+export default function(step) {
```
martypowell: this is kinda silly, not sure what i was thinking here, something to reconsider.

src/components/PaymentField.js
tmccoy529: @martypowell this has been resolved in a new branch

src/App.scss
```diff
@@ -211,3 +211,27 @@ label.tt_has-errors {
...
+
+.autocomplete-option {
+  opacity: 1;
+  padding: 5px 10px;
+}
+
+.autocomplete-results {
+  background: white;
+  box-shadow: 0 2px 12px rgba(0, 0, 0, 0.1);
+  opacity: 1 !important;
```
theilman7: Just double checking if `!important` is necessary on this. I only see one other `opacity` property and it is also set to 100% (1).

src/App.scss
```diff
@@ -211,3 +211,27 @@ label.tt_has-errors {
     background: yellow;
   }
```
theilman7: These are purely formatting items, not blocking by any means. 

• Line 51. Add spacing above the classes. Check indentation as well.

• Line 190. Remove the extra space to tighten up formatting.

src/App.scss
```diff
@@ -211,3 +211,27 @@ label.tt_has-errors {
...
+  opacity: 1;
+  padding: 5px 10px;
+}
+
+.autocomplete-results {
+  background: white;
+  box-shadow: 0 2px 12px rgba(0, 0, 0, 0.1);
+  opacity: 1 !important;
+  position: fixed;
+  z-index: 99999;
```
theilman7: For z-index, I don't see any other established values for other classes that have a z-index so I'm wondering if we should start with a smaller number vs jumping right to 99999. If the project will contain multiple z-index values, I would also recommend using SASS variables to keep things organized. If this is pursued, I would start by establishing what the highest value and corresponding elements should be, so that no other element would appear in front of it (menu overlays etc.). From there, I would work backwards to develop levels of content and apply descending values to them. For example, `$top-tier: 900;`, `$middle-tier: 600;`, `$bottom-tier: 300;`. This will ensure we don't end up with unintentional overlap down the road.

src/App.scss
```diff
@@ -211,3 +211,27 @@ label.tt_has-errors {
...
+.is-highlighted {
+  background: #e0e0e0;
+}
+
+.address-input-wrapper {
+  position: relative;
+}
+
+.autocomplete-input {
+  padding-left: 30px !important; /* account for the search icon to the left of the input text*/
```
theilman7: Double checking if this can be achieved without using `!important`.

src/App.scss
```diff
@@ -211,3 +211,27 @@ label.tt_has-errors {
...
+}
+
+.autocomplete-option {
+  opacity: 1;
```
theilman7: As the `opacity` property is usually 1 by default, wondering if we need this here. I didn't see any other `opacity` property in the file, so if this is part of a `transition` in another stylesheet and is needed, disregard.

src/components/PaymentField.jsx
```diff
@@ -0,0 +1,55 @@
...
+  form, // also values, setXXXX, handleXXXX, dirty, isValid, status, etc.
+  ...props
+}) => {
+  const { name } = field;
+  const { className, isNegativeValue, label } = props;
+  const { setFieldValue } = form;
+  const [value, setValue] = useState();
+  const cssClasses = classnames("tt_form-group flex-end total", className);
+
+  const handleChange = formattedNumber => {
```
sgrg1: it is returning Received NaN for the `value` attribute when removed the entered value in that currency field

martypowell: @sgrg1 can you please take a look at this and try to resolve it?

sgrg1: ok

src/components/PaymentField.jsx
```diff
@@ -18,8 +18,8 @@ const CustomInputComponent = ({
 
   const handleChange = formattedNumber => {
     const { valueAsNumber } = formattedNumber.target;
-    setValue(Math.abs(valueAsNumber));
-    setFieldValue(name, valueAsNumber);
+    setValue(!valueAsNumber ? undefined : Math.abs(valueAsNumber));
```
martypowell: I think the condition we want here is `0 or Math.abs(valueAsNumber)` so that it's consistent line 22?

src/common/Calculations.js
```diff
@@ -41,9 +41,8 @@ const GetCalculatedTotals = (fields = {}, monthsLate = 0) => {
     nonTransientRentalCollected,
     governmentExemptRentalCollected
   } = fields;
-
-  const totalExemptions =
-    nonTransientRentalCollected + governmentExemptRentalCollected;
+  
+  const totalExemptions = nonTransientRentalCollected+ governmentExemptRentalCollected;
```
martypowell: @sgrg1 looks like this changed, but shouldn't have.

src/components/PaymentField.jsx
tmccoy529: @martypowell the problem with changing it to 0 from undefined or empty string is that it then places a 0 in the field and you can then get the scenario of 01235.78 again which is why we removed the default 0 in the first place.

martypowell: `null` then please

sgrg1: null gives us console error
"Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
    in input (at CurrencyInput.jsx:8)
    in div (at CurrencyInput.jsx:6)"
that is why i am using undefined, which I also mention to Tim regarding this issue

martypowell: @sgrg1 something to doesn't seem right, here, but i won't block, since this control is kinda weird anyway.

src/common/Calculations.js
sgrg1: I was also playing around with this .js file may be it is showing changes but I did clean up my console.log on this page but did not change any code

martypowell: @sgrg1 it may seem like the code didn't change, but it did, if you look closely the formatting was removed, and the addition symbol is touching the nontransient variable. This is something that should be done by pretteir when saving. Do you have auto save on?

sgrg1: i resaved it and seems like formatting is working. please check

src/components/ReturnDateSelector.jsx
```diff
@@ -126,6 +134,7 @@ const ReturnInterval = props => {
                     onChange={date => handleDateChange(date, monthIndex)}
```
sgrg1: When date is removed from date control , it errors out . Need to check for if (date !== null) in HandleDateChange to fix this issue

martypowell: @sgrg1 can you hop on this too please?

sgrg1: just push the fix for this issue

src/components/PaymentField.jsx
```diff
@@ -0,0 +1,55 @@
...
+    <div className={cssClasses}>
+      <label>{label}</label>
+      <div className="tt_currency-pickers">
+        <div className="tt_currency-picker">
+          <CurrencyInput
+            isNegativeValue={isNegativeValue}
+            id={name}
+            name={name}
+            onChange={handleChange}
+            value={value}
```
sgrg1: Warning: A component is changing an uncontrolled input of type number to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component.   Solution : value={value || ''}

src/components/forms/IdentificationForm.jsx
```diff
@@ -0,0 +1,120 @@
...
+    activeStep,
+    label,
+    formik
+  } = props;
+  const { monthlyData } = formik.values;
+  const showTradeAlias = HasAtLeast1Exemption(monthlyData);
+
+  return (
+    <Formik
+      initialValues={{ email: "", nameOfSubmitter: "", titleOfSubmitter: "" }}
```
sgrg1: Warning: A component is changing an uncontrolled input of type number to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. Solution : tradeAlias:'' in initialValues

sgrg1: fixed

src/components/PaymentField.jsx
sgrg1: fixed

PR #147, opened 2019-12-13
Cleanup Paging Files before merge to integration

martypowell:
- Remove Section Files that weren't being used
- Updated form file names to standard format


src/components/forms/BasicInformationForm.jsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import { Formik, Form } from "formik";
-import TransientTaxTabs from "../../components/TransientTaxTabs";
+import TransientTaxTabs from "../TransientTaxTabs";
```
martypowell: Not blocking, but a candidate for refactoring, the implementation is more generic than TransientTax, so it can be named that way. And is a canidate for being put into a package.

PR #153, opened 2019-12-16
Change address validation

tmccoy529:
Added validation to determine if address is a valid address inside of GIS database. This triggers when there is no response from the endpoint. If there is a response then we can assume the address is valid and then move on. This handles anyone silly enough to side step the auto complete and manually type in their address.


src/components/forms/BasicInformationForm.jsx
```diff
@@ -19,17 +20,24 @@ const BasicInformationForm = props => {
...
       }}
       validationSchema={Yup.object({
         businessName: Yup.string()
           .transform(value => (!value ? null : value))
           .required("Required"),
-        businessAddress: Yup.string()
-          .transform(value => (!value ? null : value))
+        businessAddress: Yup.mixed()
           .required("Required")
+          .test("is-valid-address", "Please enter a valid address.", value => {
```
martypowell: param here should change from `value` to `address` or `addressValue`

martypowell: So when you are validating the address, it takes a little time, but there is no visual indication taht is something is happening, we should probably add a checking message. Maybe soemthing like "Validating address..."

tmccoy529: added

martypowell:

src/components/forms/BasicInformationForm.jsx
```diff
@@ -52,6 +83,12 @@ const BasicInformationForm = props => {
               name="businessAddress"
               label="Business Address"
             />
+            {isValidAddressMessage && (
+              <p role="alert" className="error-message">
```
martypowell: This probably should go into the error message comopnet someone, i can take a look at this real quick.

src/components/forms/BasicInformationForm.jsx
```diff
@@ -17,19 +18,50 @@ const BasicInformationForm = props => {
...
+
   return (
     <Formik
-      initialValues={{ businessName: "", businessAddress: "" }}
-      onSubmit={values => {
-        onValidSubmission(values);
+      initialValues={{
+        businessName: "",
+        businessAddress: "",
+        businessAddressParts: {}
```
sgrg1: Do we really need businessAddressParts:{} in initialValues? I don't see the use of it here.

martypowell: i can't remember why, but i know ive personally run into an issue with teh form values we are trying to capture is not specified. At a minimum, if we do include initial values, you know what fields the form is setting without having to inspect the values.

PR #154, opened 2019-12-18
Fixed Negative values for new currency input

tmccoy529:
Original currency input accepted a negative value and so all calculations were sums. New currency input didnt change values to negative, negative was just for looks but still held positive value. This caused calculations to be incorrect. Formik values are now negatives so calculations are now correct.

### DIFF
--- a/src/data/TaxReturnMapper.js
+++ b/src/data/TaxReturnMapper.js
@@ -188,8 +188,6 @@ const MapResponseDataForTaxReturn = taxReturn => {
     "nonTransientRentalCollected"
   ]);
 
-  console.log(exemptions);
-
   const netRoomRentals = sumTotals([occupancyTaxCollected, exemptions]);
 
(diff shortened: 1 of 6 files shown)


src/data/TaxReturnMapper.js
```diff
@@ -70,10 +70,10 @@ const getDateInformation = ({
...
   const dueDate = monthlyData.length
     ? new Date(
-        monthlyData[monthOfReturn].month +
-          "/01/" +
-          monthlyData[monthOfReturn].year
-      )
+      monthlyData[monthOfReturn].month +
+      "/01/" +
+      monthlyData[monthOfReturn].year
+    )
```
tmccoy529: Im afraid my prettier isnt working and formatting is being messed up

martypowell: i can't tell for sure, but i think this cahnge is actually right.

Issue #161, opened 2019-12-19
Input lengths 


martypowell:
Lengths for most of the inputs are off, this should probably be handled on an individual basis.

martypowell:
on the issues related to his is the currency inputs and their label wrapping..

![image](https://user-images.githubusercontent.com/1933400/71182686-82744580-2244-11ea-85a8-feb7e237c4a2.png)

PR #167, opened 2019-12-20
Fix input widths

martypowell:
Addresses #161 

Adjusts styles so form looks better. 

- Adjusted text width on business panels
- Mobile changes
- Exemption list cleanup
- Add ability to hide a label note on small screens.

### DIFF
--- a/src/components/PaymentTotal.jsx
+++ b/src/components/PaymentTotal.jsx
@@ -8,8 +8,6 @@ const PaymentTotal = props => {
   const { label, total = 0, name, className } = props;
   const cssClasses = classnames("tt_form-group total", className);
 
-  console.log(label);
-
   return (
     <div className={cssClasses}>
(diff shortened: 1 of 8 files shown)


src/components/PaymentTotal.jsx
```diff
@@ -1,15 +1,18 @@
...
 import PropTypes from "prop-types";
+import React from "react";
 import classnames from "classnames";
-import PaymentTotalLabel from "./PaymentTotalLabel";
 
 const PaymentTotal = props => {
   const { label, total = 0, name, className } = props;
   const cssClasses = classnames("tt_form-group total", className);
 
+  console.log(label);
```
tmccoy529: need to remove this logging

martypowell: @tmccoy529 on it.

PR #169, opened 2019-12-20
fixed bad merge that caused bad nesting

martypowell:
Introduced in #167 , had bad nesting causes style not to be applied (a good case against these megga sass files)


martypowell:
tested the css in staging, to verify

PR #178, opened 2019-12-23
fix quarterly console error#177

sgrg1:
fix issue with elements with non-unique id for currency input

### DIFF
--- a/src/components/PaymentField.jsx
+++ b/src/components/PaymentField.jsx
@@ -13,7 +13,8 @@ const CustomInputComponent = ({
   ...props
 }) => {
   const { name } = field;
-  const { className, isNegativeValue, label,month } = props;
+  const { className, isNegativeValue, label } = props;
+  const month = props.date.getMonth();
   const { setFieldValue } = form;
   const [value, setValue] = useState();
   const cssClasses = classnames(
@@ -40,7 +41,7 @@ const CustomInputComponent = ({
         <div className="tt_currency-picker">
           <CurrencyInput
             isNegativeValue={isNegativeValue}
-            id={`${name}_${month}`} 
+            id={`${name}-${month}`} 
             name={name}
             onChange={handleChange}
             value={value || ""}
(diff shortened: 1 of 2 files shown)


src/components/PaymentField.jsx
```diff
@@ -13,7 +13,7 @@ const CustomInputComponent = ({
   ...props
 }) => {
   const { name } = field;
-  const { className, isNegativeValue, label } = props;
+  const { className, isNegativeValue, label,month } = props;
```
martypowell: looks like your formatting is off again @sgrg1 ... there should be a space between the comma and month.

src/components/PaymentField.jsx
```diff
@@ -40,7 +40,7 @@ const CustomInputComponent = ({
         <div className="tt_currency-picker">
           <CurrencyInput
             isNegativeValue={isNegativeValue}
-            id={name}
+            id={`${name}_${month}`}
```
martypowell: can be consistent and make this `${name}-${month}` please?

src/components/PaymentField.jsx
martypowell: ok, we are actually already getting date, so we should use that, not sure what impact removing that will have. See below for more info.

src/components/PaymentField.jsx
martypowell: you should be able to get month from the date prop. date.getMonth() should give you want you want. Then you don't have to change the props and the MonthlyPaymentFormStuff

src/components/PaymentField.jsx
sgrg1: fixed

src/components/PaymentField.jsx
sgrg1: fixed

src/components/PaymentField.jsx
```diff
@@ -14,6 +14,7 @@ const CustomInputComponent = ({
 }) => {
   const { name } = field;
   const { className, isNegativeValue, label } = props;
+  const month = props.date.getMonth();
```
martypowell: you could have destructed on `date` on line 16, but not worth blocking.

PR #180, opened 2019-12-24
fix validation error for exemption list#174 

sgrg1:
fix showing correct validation error on exemption certificate

### DIFF
--- a/src/common/ExemptionUtilities.js
+++ b/src/common/ExemptionUtilities.js
@@ -10,18 +10,14 @@ const GetExemptionFormErrors = (exemption, exemptionErrorCheck=true) => {
     if (!type) {
       activeFormErrors.push({ key: "type", error: "Exemption Type Required" });
     }
-  
     if (!fromDate) {
       activeFormErrors.push({ key: "fromDate", error: "From Date Required" });
     }
-  
     if (!toDate) {
       activeFormErrors.push({ key: "toDate", error: "To Date Required" });
     }
-  
   }
   else{
-
     if (!type) {
       activeFormErrors.push({ key: "type", error: "Exemption Type Required" });
     }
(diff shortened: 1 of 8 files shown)


src/common/ExemptionUtilities.js
```diff
@@ -2,22 +2,28 @@
  * Quick Solution for Validating Exemption Errors Outside of Formik
  * @param {object} exemption object representation of an exemption
  */
-const GetExemptionFormErrors = exemption => {
+const GetExemptionFormErrors = (exemption, exemptionErrorCheck=true) => {
```
martypowell: looks like your formatting is off here.

sgrg1: fixed

martypowell: I don't understand what `exemptionErrorCheck` means.

sgrg1: Please give me the name of variable what ever make sense to you

martypowell: per our chat, i don't understand the code so I can't give you a name for the variable, but i'm not sure this change will persist after this is resolved. So i wouldn't get too hung up on this.

martypowell:
ok @sgrg1 this first step to this will be to move the [ExemptionList](https://github.com/baltimorecounty/react-transient-tax/blob/integration/src/components/ExemptionCertificateField.jsx#L58) out of the `ExemptionCertificateField` and into the `ExemptionCertificateForm`.

I would expect that to be put here.

image

This will allow us to remove the complexity of having the list and the form together, and make it easier to fix the validation the formik way.

This should be submitted as a pr independent of this one.

If you need any help doing this, or have questions, please don't hesitate to ask.

martypowell:
closing in favor of #187

Issue #171, opened 2019-12-23
Add routing for next and previous buttons


martypowell:
User should be able to navigate the form using the browser back and forward buttons

PR #181, opened 2019-12-26
Refactor Return Date Selector into Formik Field

martypowell:
Refactor work for #171 

This will allow us to pass formik values to the component after the form is submitted. This is required when you navigate forward or backward using the browser buttons.

- Changes Return Date Selector into a Formik Field Component
- Simplifies return selector logic, to two state values (months, returnStatus)
- Adds a display component for displaying the Return Date Status
- Fixes the display of dueDate (pass as date) format when it's shown


src/components/ReturnDateSelector.jsx
```diff
@@ -1,176 +0,0 @@
...
-      if (!isMonthly && isFirstMonthInQuarter) {
-        const nextMonth = addMonths(date, 1);
-        const finalMonthInQuarter = addMonths(date, 2);
-
-        newMonths[1] = nextMonth;
-        newMonths[2] = finalMonthInQuarter;
-        lastFilingMonth = finalMonthInQuarter;
-      }
-
-      const monthlyData = Object.keys(newMonths).map(monthKey => {
```
martypowell: This was missed in the changes, we need to add this.

martypowell: this was updated.

PR #183, opened 2019-12-27
Add step routing

martypowell:
**Major Refactor**

Depends on the merging of #181 and #182 

Adds ability to use browser buttons to navigate the form. This requires the following updates (sorry I know this is a big one):

- Refactor any fields we could to be more formik-y (#181, #182)
- Adds routing for steps, this requires a change to load each step dynamically, and populate the intial values based on the parent form (MultiplePageForm component)
- Adds a prompt if the user navigates the form, but there is unsaved data to ensure user doesn't lose form data
- Moves `initialValues` declaration from Forms to the step definition, this allows us to pull the values from the parent from if the exist. This is still not perfect for some values (monthly data, being the biggest issue)
- Other refactors


src/App.js
```diff
@@ -36,11 +36,14 @@ setConfig(configValues);
...
   <div className="tt_app">
     <Router>
+      <Route exact path="/">
+        <Redirect to="/steps/1" />
+      </Route>
       <Route
         exact
-        path="/"
+        path="/steps/:activeStepNumber"
         render={props => (
```
tmccoy529: Im not convinced this naming convention is the best. If a user looks at the url and sees /steps/4 and they are like 6 clicks in it might be confusing. Also if we are trying to stay consistent with BatlCoGo then this should probably be: 
/steps/BasicInformation 
/steps/PaymentInterval 
/steps/January Tax Return
/steps/Exemptions
/steps/Identification
/steps/Review

And we can probably even remove the "steps" altogether.

I just think this can cause confusion if anyone actually pays attention to the URL as its changing.

sgrg1: I believe it makes sense what Tim is saying.

src/components/PromptIfDirty.jsx
```diff
@@ -0,0 +1,15 @@
...
+import React from "react";
+import { useFormikContext } from "formik";
+
+const PromptIfDirty = props => {
+  const formik = useFormikContext();
+  return (
+    <Prompt
+      when={formik.dirty && formik.submitCount === 0}
+      message="Are you sure you want to leave? You have with unsaved changes."
+    />
```
tmccoy529: Can we modify this message to say "Are you sure you want to leave? You have unsaved changes."

martypowell: yep.

src/App.js
martypowell: we can try and use the id's but the user (shouldn't) be concerned about the url. If they for instance copy "/steps/January Tax Return" one of two things will happen,

- If it's cop and pasted when the form is active the user will be navigated to that page (but i doubt someone would do that).
- If there is no form data, the user would be redirected to the first step.

readable urls make a lot of ssense if you have something like

`/persons/martypowell` where the link will be shared. I am not sure they matter much in a form like this. I do see how it could be confusing if the user was looking at the progress bar and the urls, but with the shortened progress bar, im not sure how to get around that.

Let me try switching this to use id's rather than step number

src/components/PromptIfDirty.jsx
martypowell: updated.

src/components/forms/BasicInformationForm.jsx
```diff
@@ -6,59 +6,53 @@ import React, { useState } from "react";
...
+        businessAddress: Yup.string().required(
+          "A valid Baltimore County address is required"
+        )
       })}
     >
       {props => (
         <Form>
+          <PromptIfDirty />
           <div className="tt_form-section">
             <Field
```
tmccoy529: Do we need to check the first step for Dirty? You cant move on until its all filled in and Yup will handle the validations on that. Will this ever actually trigger?

martypowell: yes, you could theoretically change the value of  the form, and select the "Forward" browser button.

tmccoy529: Okay that makes sense

src/App.js
martypowell: @tmccoy529 I have updated this, I am still concerned about remove `/step/:stepId`. "Step" is acting essentially as a namespace for the Multiple Paged Forms

image


In the case above, someone could make a "step" with the value of "confirmation" and it would go all out of whack.

I am not opposed changing "step" to another name if you can think of something better.

tmccoy529: @martypowell I'll test this but the more I look at the code the more I see how this is going to affect the entire app and Im not sure there is enough value to make this a readable URL. I'll let you know what I find.

tmccoy529: @martypowell so this looks good. I think leaving /steps/ in is totally fine and makes sense. Bookmarking or copy and pasting the URL does throw an error.

image

If you click this link: http://localhost:3000/#/steps/payment-form-january2019 you will get the error. 
So if a user bookmarks or copies the URL and tries to open in a new browser window they will get that error.

martypowell: looking at this now

tmccoy529: @martypowell this looks good now.

martypowell:
@tmccoy529 negative value issue addressed in latest push. Steps to recreate

1. Add exemption to monthly payment
2. Go to next page
3. Go back, without fix, the number was negative.

Issue #99, opened 2019-11-25
Why do I need this content needs to be populated.


martypowell:
We need to provide content for the following "Why do I need this?" buttons (email included)

![image](https://user-images.githubusercontent.com/1933400/69544672-3512fa00-0f5e-11ea-9758-b2b5a40ab915.png)

martypowell:
@sdurphy can i pawn this off to you?

sdurphy:
@martypowell sure you can.

sdurphy:
@martypowell please us the following copy approved by OBF:

Return Submitted By:
We collect this information in order to keep record of who will be submitting the payment. 

Title:
We collect this information to determine how the submitter is affiliated with the business.

martypowell:
@sdurphy we have one for `Email`. This is what we have...

"We will send you a confirmation number that you will bring when you pay the form in person."

Is this okay?

sdurphy:
@martypowell can you change to the following for now? (I haven't run this by Heather L yet):

"We will send a confirmation number to the email address you provide. Please bring this confirmation number with you when you submit your payment in person. "

PR #184, opened 2019-12-30
Updated content for why do i need this

martypowell:
Addresses #99 

Adjusts tests per recommendation.


martypowell:
@tmccoy529 just fyi I was testing this out in draft mode, so it's possible more changes would come, in general i think it would be good to leave review until the is "officially" submitted.

PR #186, opened 2019-12-30
Added months late to props

tmccoy529:
Fixed an issue where penalties were not being calculated because monthsLate was not being added to the props and were always 0.


src/components/hooks/useReturnInterval.js
```diff
@@ -33,11 +33,13 @@ const useReturnInterval = ({
     if (isIntervalSelected) {
       const lastFilingMonth = months[Object.keys(months).length - 1];
       const returnStatus = GetDueDateStatus(lastFilingMonth, new Date());
+      const { value, isLate } = returnStatus;
```
martypowell: yea, that looks great.

PR #190, opened 2019-12-31
Added tests for Critical Calculation function

martypowell:
There were not tests to cover the calculation function that is used on the review and confirmation panels. Added test to cover the possible scenarios, and refactored logic to be more testable.

Also removed an unused file.


martypowell:
@baltimorecounty/front-end-developers I will note, the function effected, could probably use a refactor to pass the params as an object instead of 5 individual values, but I didn't want to take the risk of impacting the functionality at this point.

martypowell:
@tmccoy529 @sgrg1 this one needs a look please.

Issue #197, opened 2019-12-31
Must add a date error message hangs around


tmccoy529:
On Payment Selection Page:

Select a filing type then click the next button. You will see an error message display. Now select a date and the error message stays around until you click next again.

![image](https://user-images.githubusercontent.com/40430541/71630264-0547b980-2bd0-11ea-8e16-e4c64a87eb96.png)

PR #204, opened 2020-01-02
Addresses Payment Interval Validation Hanging Around

martypowell:
Addresses #197 

This mostly resolves this issue, I found some cases if you are continually switching stuff in the form where the message shows when it probably shouldn't. But when a value is provided, the message no longer shows, which is the most important.

There was an issue with setting multiple form values inside a field, so they had to be consolidated into a single object. These changes do that, and any repercussions to that change.

### DIFF
--- a/src/common/Calculations.js
+++ b/src/common/Calculations.js
@@ -35,7 +35,7 @@ const CalculateTaxCollected = netRoomRentalCollections =>
  * @param {number} monthsLate number of months tardy on payment, this will determine if penalty and interest is charged
  * @returns {object} an object containing all the desired totals
  */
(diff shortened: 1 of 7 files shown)


tmccoy529:
@martypowell found an issue in this where exemption certificate screen appears when its not supposed to.

Select quarterly type. Added exemption for first month, click next and you will go to exemption certificate screen. Add a cert and click next will then go to next month of the quarterly return.

martypowell:
> @martypowell found an issue in this where exemption certificate screen appears when its not supposed to.
> 
> Select quarterly type. Added exemption for first month, click next and you will go to exemption certificate screen. Add a cert and click next will then go to next month of the quarterly return.

ok @tmccoy529 I'm seeing this too, addressing now.

PR #210, opened 2020-01-02
fix issue with next click on exemption edit#194

sgrg1:
Don't let user go to next page while exemption list is in edit (update) mode.


src/components/forms/ExemptionCertificateForm.jsx
```diff
@@ -23,7 +23,10 @@ const ExemptionCertificateForm = props => {
     <Formik
       initialValues={initialValues}
       onSubmit={values => {
-        onValidSubmission(values);
+        const { isExemptionFormDirty } = values;
+        if ( isExemptionFormDirty || isExemptionFormDirty === undefined ) {
```
martypowell: I think we should set the default value here of "isExemptionFormDirty " to false instead of 0 here, https://github.com/baltimorecounty/react-transient-tax/blob/5e1988ca7bf684b5a864ae8da3fd855bbd26c63d/src/components/formik/ExemptionCertificateField.jsx#L19.

This should probably happen regardless of the change above.

then this check here can just be `if (isExemption)` ???

martypowell: After looking into this more, i've found more issues. I will report back as soon as I can summarize them.

martypowell: fyi @sgrg1

martypowell: Okay part one, see the comments in the image to see the issues.

image

src/components/PromptIfUpdating.jsx
```diff
@@ -0,0 +1,16 @@
+import { Prompt } from "react-router-dom";
+import React from "react";
+import { useFormikContext } from "formik";
+
+const PromptIfUpdating = props => {
```
martypowell: This is kinda what I expected, but why duplicate the component?

I think you could use the existing component, and the when condition could check for either isDirty.

You can then add a message prop, and use the existing message for PromptIfDirty as the default prop value. And then update the [line](https://github.com/baltimorecounty/react-transient-tax/pull/210/files/07579971cd1b03337d8064dbcd79a592198cf474..f4413e6d951d3c80a6a52df78f8e81bb285948d7#diff-023c6641fa52b1ce7dcaa76835cfb6f9R43) for the message in this component.

Does that make sense?

src/components/PromptIfDirty.jsx
```diff
@@ -4,11 +4,19 @@ import { useFormikContext } from "formik";
 
 const PromptIfDirty = props => {
   const formik = useFormikContext();
+  const { isExemptionFormDirty } = formik.values;
```
martypowell: formatting is off here.

src/components/PromptIfUpdating.jsx
martypowell: Sounds like this panel might be going away anyway, so approving as is.

Issue #102, opened 2019-11-25
Populate error page content


martypowell:
There is an error page that we redirect to if the app encounters any errors, but we need to populate this.

martypowell:
@sdurphy can I dump this on you please?

sdurphy:
@martypowell will this be an alternate view of the confirmation page? (Ex: Tax Sale) And if so, do we have the ability to spit out a description of what the error is on that page?

martypowell:
@sdurphy I think for this case, we can assume something went wrong, types of errors

- Network Error - Service is not available
- Anything else - maybe we want to asay check again in a couple of minutes, you can still file in person or something like that?

PR #215, opened 2020-01-03
Error page

martypowell:
Addresses #102 

- Added check to the app to verify the api is up, if it's not, then we will redirect them to an error page, because they won't be able to submit a form or check their confirmation number.
- Refactored some of the repeated logic mainly found in the error page

I tried to put this network check in app.js and couldn't get it to work, so for the sake of the time, I'm pushing this through with the repeated code, it is probably something we should come back to, and see if we can handle this in one place.


martypowell:
@sgrg1 @tmccoy529 I am waiting for @sdurphy to get some feedback on one of the messages. This won't be anything but a string change so I will still merge after that's done.

PR #222, opened 2020-01-08
fixed interest again for all forms of review

martypowell:
Last change fixed review and monthly panel but did not account for the confirmation page, where we have to calculate this again. So i rolled back so that all review pieces have to use the calculation.

### DIFF
--- a/src/common/DatesUtilities.test.js
+++ b/src/common/DatesUtilities.test.js
@@ -102,6 +102,19 @@ describe("Get Due Date Status", () => {
       })
     );
   });
+
+  test("should days remaining status, when the date of filing is before than the due date by 1 day", () => {
+    const actual = GetDueDateStatus(
+      new Date("November 30, 2019"),
+      new Date("2020-01-08T14:36:04")
+    );
+    expect(actual).toEqual(
+      expect.objectContaining({
+        isLate: true,
+        value: 1
+      })
+    );
+  });
 });
 
 describe("Get Min Date", () => {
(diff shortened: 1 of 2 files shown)


src/common/DatesUtilities.test.js
```diff
@@ -102,6 +102,19 @@ describe("Get Due Date Status", () => {
       })
     );
   });
+
+  test("should days remaining status, when the date of filing is before than the due date by 1 day", () => {
```
martypowell: this can be removed.

src/data/TaxReturnMapper.js
```diff
@@ -126,21 +127,18 @@ const GetReturnSummaryValues = taxReturnValues => {
  * Maps Transient Tax Form Data to the confirmation page
  * @param {object} taxReturn tax return form data model
  */
-const MapResponseDataForTaxReturn = taxReturn => {
+const MapResponseDataForTaxReturn = (taxReturn, isServerData = false) => {
```
martypowell: isServerData is not being used.

src/data/TaxReturnMapper.js
```diff
@@ -164,6 +162,8 @@ const MapResponseDataForTaxReturn = taxReturn => {
     CalculateTaxCollected(netRoomRental)
   );
 
+  monthlyData.map(({ interestRemitted }) => interestRemitted);
```
martypowell: this should be rolled back

PR #223, opened 2020-01-09
Fix Positive Number Exemption

martypowell:
The confirmation page was displaying exemptions as a positive number because that is what the api was returning. 

This fix that when mapping values, the exemptions will always be negative.

### DIFF
--- a/src/services/ApiService.js
+++ b/src/services/ApiService.js
@@ -49,7 +49,7 @@ const GetTransientTaxReturn = confirmationNumber =>
       )}/transientTax/return?confirmationnumber=${confirmationNumber}`
     )
     .then(({ status, data }) =>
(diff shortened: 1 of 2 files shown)


src/services/ApiService.js
```diff
@@ -49,7 +49,7 @@ const GetTransientTaxReturn = confirmationNumber =>
       )}/transientTax/return?confirmationnumber=${confirmationNumber}`
     )
     .then(({ status, data }) =>
-      status === 200 ? MapResponseDataForTaxReturn(data) : []
+      status === 200 ? MapResponseDataForTaxReturn(data, true) : []
```
martypowell: this shouldn't have made it in here.

PR #224, opened 2020-01-13
Remove date min max

tmccoy529:
This PR is to remove forcing users to match exemption dates with monthly dates per user request. 

With that being removed this introduced the bug of bad date ranges where the To date could come before the From date. In order to handle that we now do a check specific for those dates (converting dates to time and comparing) and then flagging the fields with an error if they dont work.

### DIFF
--- a/src/common/ExemptionUtilities.js
+++ b/src/common/ExemptionUtilities.js
@@ -21,6 +21,15 @@ const GetExemptionFormErrors = exemption => {
     activeFormErrors.push({ key: "toDate", error: "To Date Required" });
   }
 
+  if (fromDate && toDate) {
+    if (parseInt(fromDate.getTime()) > parseInt(toDate.getTime())) {
+      activeFormErrors.push({
+        key: "badDateRange",
+        error: "To Date must come after from date"
+      });
+    }
+  }
+
   return activeFormErrors;
 };
 
--- a/src/components/formik/ExemptionSelector.jsx
+++ b/src/components/formik/ExemptionSelector.jsx
@@ -131,11 +131,13 @@ const ExemptionSelector = props => {
       {(touched.fromDate || touched.toDate) &&
         formErrors.some(
           ({ key }) => key === "toDate" || key === "fromDate"
-        ) && <BasicErrorMessage message="To Date and From Date are Required" />}
+        ) && (
+          <BasicErrorMessage message="To and From Date are required to submit an exemption." />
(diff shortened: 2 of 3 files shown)


src/common/ExemptionUtilities.js
```diff
@@ -21,6 +21,15 @@ const GetExemptionFormErrors = exemption => {
     activeFormErrors.push({ key: "toDate", error: "To Date Required" });
   }
 
+  if (fromDate && toDate) {
+    if (parseInt(fromDate.getTime()) > parseInt(toDate.getTime())) {
+      activeFormErrors.push({
+        key: "badDateRange",
```
tmccoy529: I saw mixed ideas online if getTime returned an integer only so I wrapped in parseInt to ensure we always are comparing integers. This might be over kill and can probably be removed.

src/components/formik/ExemptionSelector.jsx
```diff
@@ -134,15 +127,18 @@ const ExemptionSelector = props => {
...
-          ({ key }) => key === "fromDate" || key === "toDate"
+          ({ key }) => key === "toDate" || key === "fromDate"
         ) && (
           <BasicErrorMessage message="To and From Date are required to submit an exemption." />
         )}
+      {touched.fromDate &&
+        touched.toDate &&
+        formErrors.some(({ key }) => key === "badDateRange") && (
+          <BasicErrorMessage message="To Date must come after From Date." />
+        )}
```
tmccoy529: In the beginning these two were merged to look lke  ({ key }) => key === "toDate" || key === "fromDate" || key ==="badDateRange" but the message wasnt generic enough to handle a bad date range so I added an additional catch here. If we can come up with a better message that can encompass both scenarios of a missing date or a bad date range then these could be easily merged.

src/common/ExemptionUtilities.js
tmccoy529: Also the error message here is confusing. We never use this message that I can tell. We may want to consider just sending key down the pipe instead of both.

martypowell:
this is being kept open, because we want to explore displaying the validation message.

martypowell:
@tmccoy529 noticed this was open, do we still think we should keep this open?

tmccoy529:
@martypowell I see no reason to keep this.

martypowell:
Closed per discussion with Tim

PR #227, opened 2020-01-15
Clear up confirmation

martypowell:
- Accounts for service changes by changing the address parts field
- Adds business name to confirmation page

@tmccoy529 @sgrg1 if you test this, you will need to get the latest from the service

### DIFF
--- a/src/components/formik/AddressLookupField.jsx
+++ b/src/components/formik/AddressLookupField.jsx
@@ -51,12 +51,10 @@ const CustomInputComponent = ({
     }
   };
 
-  const handleAddressSelect = (value, item, test) => {
+  const handleAddressSelect = (value, item) => {
     setFieldValue("businessAddress", value);
     setFieldValue("address", item);
     setWasSelected(true);
-
-    console.log(value, item, test);
   };
 
   /**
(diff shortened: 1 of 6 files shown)


src/components/formik/AddressLookupField.jsx
```diff
@@ -47,14 +47,16 @@ const CustomInputComponent = ({
...
     } else {
       setFieldValue("businessAddress", value);
-      setFieldValue("businessAddressParts", {});
+      setFieldValue("address", {});
     }
   };
 
-  const handleAddressSelect = (value, item) => {
+  const handleAddressSelect = (value, item, test) => {
     setFieldValue("businessAddress", value);
```
tmccoy529: what is "test" doing?

src/components/formik/AddressLookupField.jsx
```diff
@@ -47,14 +47,16 @@ const CustomInputComponent = ({
...
-    setFieldValue("businessAddressParts", item);
+    setFieldValue("address", item);
     setWasSelected(true);
+
+    console.log(value, item, test);
   };
```
tmccoy529: need to remove console.log

src/components/formik/AddressLookupField.jsx
martypowell: shoudl be gone, good catch.

src/components/formik/AddressLookupField.jsx
martypowell: yes

PR #230, opened 2020-01-30
Sum Totals After They are calculated

### LINKED ISSUE #229: Rounding Issue with Calculations for Quarterly Returns
Per @tmccoy529 

While helping Sanjay trouble shoot a value difference I came across a sort of unique situation where we are getting different values that are both correct based on displayed values and might cause some confusion. It has to do with decimal places and rounding and math.

So down below we have our interest calculations on a return that starts in March of 2019

March: 1000 collected
                300 exempt
April: 2000 collected
                300 exempt
May: 3000 collected
                300 exempt

March Interest Calculation:

66.5 x .07 = 4.655 rounded to 4.66 shown below

April Interest Calculation:

161.5 x .07 = 11.305 rounded to 11.31 shown below

May Interest Calculation:

256.5 x .07 = 17.955 rounded to 17.96 shown below

Total Un-Rounded:
33.915

Total Rounded:
33.93

Total Interest Calculation:
484.5 x .07 = 33.915 rounded to 33.92

There is a penny discrepancy if you add all the rounded interest values up versus the correct value of the rounded total value. The county still gets the correct amount of money but doing the calculations manually based on what we are displaying the tax payer is shorting the county one cent. We may need to display more decimal places then 2 to handle this issue but I feel like that is something OBF would have to decide. Or maybe it’s not a big deal since it’s just a penny and we can add legal text to explain math and rounding. Anyway figured I bring it to everyone’s attention incase this was found post release and was deemed an issue.

image

martypowell:
Resolves #229 

Total Values were being run through calculation functions after the individual months had already been calculated. 

We only want to run the calculations on the individual months, and then add the calculated values.  This ensures the numbers match up and avoids the scenario in #229.

### DIFF
--- a/src/common/Calculations.js
+++ b/src/common/Calculations.js
@@ -1,5 +1,7 @@
 import { RatesAndFees } from "./Constants";
 
+const roundCurrency = value => Math.round(value * 100) / 100;
+
 /**
  * Calculate the interest based on tax collected and the number of months late.
  * @param {number} taxCollected dollar amount to base the interest off
@@ -10,7 +12,7 @@ const CalculateInterest = (
   taxCollected,
   numberOfMonthsLate,
   interestRate = RatesAndFees.InterestRate
-) => taxCollected * numberOfMonthsLate * interestRate;
+) => roundCurrency(taxCollected * numberOfMonthsLate * interestRate);
 
 /**
  * Calculate the penalty fee based on tax collected.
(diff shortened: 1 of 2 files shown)


martypowell:
@tmccoy529 @sgrg1 I am going to push this up, now, so that it's available for the customer, but will still treat this like a normal pr.

sgrg1:
@martypowell , I think our calculation for interest is still bad . I am testing this branch with following value.
buggy

I believe interest should be around $11.30.  Issue was rounding the number

martypowell:
> @martypowell , I think our calculation for interest is still bad . I am testing this branch with following value.
> buggy
> 
> I believe interest should be around $11.30. Issue was rounding the number

will checkout now.

martypowell:
> Sorry, thought this had been approved last week

Me too, @sgrg1 found something, but i believe the nubmers changed because it was the end of the week.

PR #231, opened 2020-02-03
fix exemption message for Non-Transient

sgrg1:
please view the issue 15935 in the system of record.

If Non-Transient is selected, then the ‘From’ and ‘To’ date range must be at least 90 consecutive days regardless of type of return (Monthly/Quarterly) or month range of return (e.g., Jan-Mar 2019), i.e., any 90 day or greater range is acceptable. 
 

Also, the word “from” in the Non-Transient description should be changed to “for”.

### DIFF
--- a/src/common/DatesUtilities.js
+++ b/src/common/DatesUtilities.js
@@ -105,7 +105,7 @@ const GetMaxExemptionEndDate = (monthlyData = []) => {
   return endOfMonth(new Date(year, month - 1, 1));
 };
 const IsDateInRange = (fromDate, toDate) =>
-  differenceInDays(toDate, fromDate) >= 90 ? true : false;
+  differenceInDays(toDate, fromDate) >= 90;
 
 export {
   DefaultDateFormat,
--- a/src/common/DatesUtilities.js
+++ b/src/common/DatesUtilities.js
@@ -104,6 +104,22 @@ const GetMaxExemptionEndDate = (monthlyData = []) => {
   const { month, year } = monthlyData.filter(hasExemption).pop();
   return endOfMonth(new Date(year, month - 1, 1));
 };
+const isDateInRange = (fromDate, toDate) => {
+  var one_day = 1000 * 60 * 60 * 24;
+
+  // Convert both dates to milliseconds
+  var date1_ms = fromDate.getTime();
+  var date2_ms = toDate.getTime();
+
+  // Calculate the difference in milliseconds
+  var difference_ms = date2_ms - date1_ms;
+
+  // Convert back to days and return
+  let totalDays = Math.round(difference_ms / one_day);
+
+  let isDaysValid = totalDays >= 90 ? true : false;
+  return isDaysValid;
+};
 
 export {
   DefaultDateFormat,
@@ -112,5 +128,6 @@ export {
   GetDueDateStatus,
   GetFormatedDateTime,
   GetMinExemptionStartDate,
-  GetMaxExemptionEndDate
+  GetMaxExemptionEndDate,
+  isDateInRange
 };
--- a/src/common/DatesUtilities.js
+++ b/src/common/DatesUtilities.js
@@ -104,7 +104,7 @@ const GetMaxExemptionEndDate = (monthlyData = []) => {
   const { month, year } = monthlyData.filter(hasExemption).pop();
   return endOfMonth(new Date(year, month - 1, 1));
 };
-const IsDateInRange = (fromDate, toDate) =>
+const IsDateRangeAtLeast90Days = (fromDate, toDate) =>
   differenceInDays(toDate, fromDate) >= 90;
 
 export {
@@ -115,5 +115,5 @@ export {
   GetFormatedDateTime,
   GetMinExemptionStartDate,
   GetMaxExemptionEndDate,
-  IsDateInRange
+  IsDateRangeAtLeast90Days
 };
--- a/src/common/DatesUtilities.js
+++ b/src/common/DatesUtilities.js
@@ -104,22 +104,8 @@ const GetMaxExemptionEndDate = (monthlyData = []) => {
   const { month, year } = monthlyData.filter(hasExemption).pop();
   return endOfMonth(new Date(year, month - 1, 1));
 };
-const isDateInRange = (fromDate, toDate) => {
-  var one_day = 1000 * 60 * 60 * 24;
-
-  // Convert both dates to milliseconds
-  var date1_ms = fromDate.getTime();
-  var date2_ms = toDate.getTime();
-
-  // Calculate the difference in milliseconds
-  var difference_ms = date2_ms - date1_ms;
-
-  // Convert back to days and return
-  let totalDays = Math.round(difference_ms / one_day);
-
-  let isDaysValid = totalDays >= 90 ? true : false;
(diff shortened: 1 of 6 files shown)


src/common/DatesUtilities.js
```diff
@@ -104,6 +104,22 @@ const GetMaxExemptionEndDate = (monthlyData = []) => {
   const { month, year } = monthlyData.filter(hasExemption).pop();
   return endOfMonth(new Date(year, month - 1, 1));
 };
+const isDateInRange = (fromDate, toDate) => {
```
martypowell: All exporeted modules should be PascalCase so this should be `IsDateInRange` (this will break the tests, so theyll need to be updated there)

Using dateFns you can simplify this function to:

```js
const isDateInRange = (fromDate, toDate) => differenceInDays(toDate, fromDate) >= 90
```
Typically if we have any date logic, try to use date-fns first.

There is one thing that I think is misleading. There is nothing in this function name that says anything about the 90 days.

I would recommend naming this function something like `IsDateRangeAtLeast90Days`. It's a bit wordy, but someone will know exactly what it's doing.

src/common/ExemptionUtilities.js
```diff
@@ -4,7 +4,7 @@
  */
 const GetExemptionFormErrors = exemption => {
   const activeFormErrors = [];
-  const { fromDate, toDate, type } = exemption;
+  const { fromDate, toDate, type, isDaysValid } = exemption;
```
martypowell: i would consider making this key `isDateRangeAtLeast90days`

martypowell: I would recommend this change throughout the other files.

src/components/formik/ExemptionSelector.jsx
```diff
@@ -60,19 +63,36 @@ const ExemptionSelector = props => {
     }
   };
 
+  const checkNonTransientDateCondition = (type, fromDate, toDate) => {
```
martypowell: this can be simplified to:

```js
const checkNonTransientDateCondition = (type, fromDate, toDate) =>
  type === 1 && fromDate && toDate
        ? isDateInRange(fromDate, toDate)
        : false
```

src/components/formik/ExemptionSelector.jsx
```diff
@@ -60,19 +63,36 @@ const ExemptionSelector = props => {
...
+    let isDaysValid = false;
+    isDaysValid =
+      type === 1 && fromDate && toDate
```
martypowell: what does type === 1  mean?

sgrg1: value of the first radio button "Non-Trasient"

martypowell: I think you have to get exemption types by 

```js
const nonTransientExemptionType = exemptionTypes.find(x => x.Description === "Non-Transient");
type === nonTransientExemptionType.Id
```

This ensures if the database values change for id, that we will still get the right match. Feel free to simplify this logic, this is just to get you in the right direction.

src/common/DatesUtilities.js
sgrg1: done

src/common/ExemptionUtilities.js
sgrg1: done

src/components/formik/ExemptionSelector.jsx
sgrg1: done

src/common/DatesUtilities.js
```diff
@@ -104,6 +104,8 @@ const GetMaxExemptionEndDate = (monthlyData = []) => {
   const { month, year } = monthlyData.filter(hasExemption).pop();
   return endOfMonth(new Date(year, month - 1, 1));
 };
+const IsDateInRange = (fromDate, toDate) =>
+  differenceInDays(toDate, fromDate) >= 90 ? true : false;
```
martypowell: This does not require a tenary operarator

```js
differenceInDays(toDate, fromDate) >= 90 ? true : false;
```

The result of this function is a boolean for example:

```js
2 > 1 
```
will always return true, just as

```js
1 > 2
```

will always return false

therefor the ` true : false` is redundant

so this can beccome

```js
differenceInDays(toDate, fromDate) >= 90
```

src/common/DatesUtilities.js
martypowell: @sgrg1 looks like you didn't update the function name to `IsDateRangeAtLeast90Days`. Note this will also break the added tests, all you will need to do is adjust the names inside the dateutility tests.

src/components/formik/ExemptionSelector.jsx
sgrg1: done

src/common/DatesUtilities.js
sgrg1: got it . modified

src/common/DatesUtilities.js
sgrg1: done

src/components/formik/ExemptionSelector.jsx
```diff
@@ -60,19 +65,40 @@ const ExemptionSelector = props => {
     }
   };
 
+  const checkNonTransientDateCondition = (type, fromDate, toDate) =>
+    type === nonTransientExemptionType.Id && fromDate && toDate
+      ? IsDateRangeAtLeast90Days(fromDate, toDate)
+      : false;
+
   const handleExemptionTypeChange = ({ type, label }) => {
+    let { fromDate, toDate } = exemption;
```
martypowell: not blocking, is there a reason you are using let here instead of `const` ?

src/components/formik/ExemptionSelector.jsx
```diff
@@ -60,19 +65,40 @@ const ExemptionSelector = props => {
...
+    );
     setExemption({
       ...exemption,
-      ...{ type, label }
+      ...{ type, label, isDateRangeAtLeast90days }
     });
   };
 
   const handleExemptionDateChange = ({ fromDate, toDate }) => {
+    let isDateRangeAtLeast90days = true;
```
martypowell: not blocikning but this coudl be simplified to

```js
const type = Object.values(exemption).length > 0 ? exemption.type : undefined;
const isDateRangeAtLeast90days = !type || checkNonTransientDateCondition(type, fromDate, toDate);
setExemption(...
```

what this says is if type is not defined, set isDateRangeAtLeast90days to true, otherwise get the Boolean value from the check function.

I would also consider this logic

```js
const type = Object.values(exemption).length > 0 ? exemption.type : undefined;
```

to be

```js
const type = Object.values(exemption).length > 0 ? exemption.type : null;
```

On the surface this may not seem like a big deal but I tend to agree with this [answer](https://stackoverflow.com/a/6604783/1143670) on stack overflow, and another good [description](https://stackoverflow.com/a/37980662/1143670). In this case, this is an intentional absence of a value.

src/components/formik/ExemptionSelector.jsx
sgrg1: done

src/components/formik/ExemptionSelector.jsx
sgrg1: done

PR #232, opened 2020-02-19
Bug#15935

sgrg1:
Fix issue with consecutive days range .

### DIFF
--- a/src/common/DatesUtilities.js
+++ b/src/common/DatesUtilities.js
@@ -6,10 +6,7 @@ import {
   endOfMonth,
   format,
   startOfDay,
-  startOfMonth,
-  getYear,
-  getMonth,
-  getDate
+  startOfMonth
 } from "date-fns";
 
 const DefaultDateFormat = "MMMM d, yyyy";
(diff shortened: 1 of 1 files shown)


src/common/DatesUtilities.js
```diff
@@ -104,8 +104,9 @@ const GetMaxExemptionEndDate = (monthlyData = []) => {
   const { month, year } = monthlyData.filter(hasExemption).pop();
   return endOfMonth(new Date(year, month - 1, 1));
 };
-const IsDateRangeAtLeast90Days = (fromDate, toDate) =>
-  differenceInDays(toDate, fromDate) >= 90;
+const IsDateRangeAtLeast90Days = (fromDate, toDate) => {
+  return differenceInDays(toDate, fromDate) + 1 >= 90;
```
martypowell: why are we adding 1 here?

sgrg1: If you have todate: 02/29/2020 and from date 02/01/2020. differenceInDays function is returning 28 days(29-1) but actually we wanted total 29 days. That is why I am adding +1 to compensate the missing day.

martypowell: can we add a comment to day that differenceInDays does not account for full days for example
jan 1 2019 to jan 31 2019 returns 30 days?

sgrg1: comment added

PR #234, opened 2020-03-18
Fix bug#16506  on Returns Submitted on the 31st

sgrg1:
Bug 16506.
it was actually calculating Penalty incorrectly.


martypowell:
@sgrg1 the bug says interest but in the overview it says penalty? is that just a typo?

sgrg1:
There was no problem with interest . Penalty had problem

martypowell:
@sgrg1 are you saying that Bug 16506 is not reproducible?

sgrg1:
I could not recreate that bug when I pass 01/31/2020 as a submitted date but I saw incorrect penalty amount for that date.

martypowell:
> I could not recreate that bug when I pass 01/31/2020 as a submitted date but I saw incorrect penalty amount for that date.

So you should update that particular bug as not able to reproduce and ensure with that the business owner confirms that is no longer an issue. Then log the bug that you found, so that the code can be reviewed based on that criteria.

PR #236, opened 2020-03-25
fix penalty calculation

sgrg1:
bugfix for #16084. penalty was not calculated correctly .


sgrg1:
test case added

martypowell:
> test case added

@sgrg1 woudln't you want to test that the penalty is also applied after two months?

sgrg1:
Added second test case for more than one month delay

PR #237, opened 2020-05-11
Bug#17153 update currency amounts

sgrg1:
Update Currency Amounts to Include Commas and Decimals

### DIFF
--- a/src/common/Calculations.js
+++ b/src/common/Calculations.js
@@ -51,7 +51,7 @@ const PreserveDecimalFormatNumber = n => {
 };
 const FormattedAmount = (value, isNegativeValue) => {
   let fieldValue = 0;
-  if (value === "") {
+  if (value.trim() === "") {
     return [0, fieldValue];
   }
 
--- a/src/common/Calculations.js
+++ b/src/common/Calculations.js
@@ -38,8 +38,56 @@ const ParseAmountToFloat = amount => {
 };
 const FormatNumber = n => {
   // format number 1000000 to 1,234,567
-  console.log("n value:" + n);
-  return n.toString().replace(/\D/g, "").replace(/\B(?=(\d{3})+(?!\d))/g, ",");
+  return n
+    .toString()
+    .replace(/^0+/, "")
+    .replace(/\D/g, "")
+    .replace(/\B(?=(\d{3})+(?!\d))/g, ",");
+};
+const PreserveDecimalFormatNumber = n => {
+  return n.toString().indexOf(",") >= 0
+    ? n
+    : n.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
+};
+const formattedAmount = (value, isNegativeValue) => {
+  let fieldValue = 0;
+  if (value === "") {
+    return [0, fieldValue];
+  }
+
+  if (value.indexOf(".") >= 0) {
+    var decimal_pos = value.indexOf(".");
+    var left_side = value.substring(0, decimal_pos);
+    var right_side = value.substring(decimal_pos);
+    if (left_side.length > 0) {
+      value =
+        FormatNumber(left_side) +
+        "." +
+        FormatNumber(right_side).substring(0, 2);
+
+      fieldValue = isNegativeValue
+        ? -ParseAmountToFloat(value)
+        : ParseAmountToFloat(value);
+    } else {
+      var rightSideLength = right_side.length;
+      var rightFormatedValue = "." + FormatNumber(right_side).substring(0, 2);
+      value = rightSideLength === 1 ? value : rightFormatedValue;
+      fieldValue =
+        rightSideLength === 1
+          ? 0
+          : isNegativeValue
+          ? -ParseAmountToFloat(rightFormatedValue)
+          : ParseAmountToFloat(rightFormatedValue);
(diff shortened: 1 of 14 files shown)


src/common/Calculations.js
```diff
@@ -33,6 +33,63 @@ const CalculateTaxCollected = (
   taxRate = RatesAndFees.TransientTaxRate
 ) => roundCurrency(netRoomRentalCollections * taxRate);
 
+const ParseAmountToFloat = amount => {
```
martypowell, line 36: These functions seem more like [FormatUtilities](https://github.com/baltimorecounty/react-transient-tax/blob/integration/src/common/FormatUtilities.js) than Calculations. It would also be good to add at least one test for each of these functions.

martypowell: Also would be nice to have at least a one line summary for each function. There are several examples of this in the common folder.

src/components/formik/PaymentField.jsx
```diff
@@ -15,22 +18,23 @@ const CustomInputComponent = ({
...
     /** IE 11 valueAsNumber does not work, so we have to use the string "value" from the target */
-    const { value, valueAsNumber } = formattedNumber.target;
-    const currencyValue = valueAsNumber || parseFloat(value);
-
-    setValue(!currencyValue ? undefined : Math.abs(currencyValue));
-    setFieldValue(
-      name,
-      !currencyValue ? 0 : isNegativeValue ? -currencyValue : currencyValue
-    );
+    let value = formattedNumber.target.value.trim();
```
martypowell, line 33: this can be cleaned up a bit.

```js
const [value, fieldValue] = FormattedAmount(formattedNumber.target.value.trim(), isNegativeValue);

setValue(value);
setFieldValue(name, fieldValue);
```

you may also want to handle the trim inside of FormattedAmount instead of here, so that you don't have to do this every time you call this function.

src/components/forms/MonthlyPaymentForm.jsx
```diff
@@ -51,8 +51,16 @@ const MonthlyPaymentForm = props => {
...
-        grossRentalCollected: Yup.number()
-          .min(0.01, "Specify an amount for gross occupancy.")
+        grossRentalCollected: Yup.string()
+          .test(`test-name`, `Specify an amount for gross occupancy.`, function(
+            value
+          ) {
+            return (value !== undefined
+              ? Number(value.toString().replace(/,/g, ""))
+              : 0) >= 0.01
+              ? true
```
martypowell, line 61: a ternary operator is not neccessary if the condition results to a true/ false value, for example

```js
const isTwo = 1 + n === 2 ? true : false;
```

is the same as

```js
const isTwo = 1 + 1 === 2;
```

the condition always returns a boolean like in your condition

```js
(value !== undefined
  ? Number(value.toString().replace(/,/g, ""))
  : 0) >= 0.01
```

src/common/FormatUtilities.js
```diff
@@ -18,4 +18,84 @@ const FormatCurrency = dollarAmount => currencyFormatter.format(dollarAmount);
...
+/**
+ * Parse string to float
+ * @param {string} amount
+ */
+const ParseAmountToFloat = amount => {
+  return parseFloat(amount.toString().replace(/,/g, ""));
+};
+/**
+ * Format enter amount to currency string
+ * @param {string} n
```
martypowell, line 30: if n is really a string you shouldn't need to convert it to a string on line 35?

src/common/FormatUtilities.js
```diff
@@ -18,4 +18,84 @@ const FormatCurrency = dollarAmount => currencyFormatter.format(dollarAmount);
...
+const FormatNumber = n => {
+  // format number 1000000 to 1,234,567
+  return n
+    .toString()
+    .replace(/^0+/, "")
+    .replace(/\D/g, "")
+    .replace(/\B(?=(\d{3})+(?!\d))/g, ",");
+};
+/**
+ * preserve currency string with deciaml when previous button is clicked
```
martypowell, line 41: This function isn't specific to a button being clicked, that just happens to be where you are using it. A summary should only talk about the function it's describing.

src/common/FormatUtilities.js
```diff
@@ -18,4 +18,84 @@ const FormatCurrency = dollarAmount => currencyFormatter.format(dollarAmount);
...
+ * @param {string} n
```
martypowell, line 42: same comment as above, is this a string? because you are calling toString below.

src/components/forms/MonthlyPaymentForm.jsx
sgrg1: done

src/common/FormatUtilities.js
sgrg1: done

src/common/FormatUtilities.js
sgrg1: done

src/common/FormatUtilities.js
sgrg1: done

src/common/FormatUtilities.js
sgrg1: done

PR #241, opened 2020-08-11
16508 trade alias address

sgrg1:
added logic to handle new field added Transaction number  and date


sgrg1:
no need to do PR on this

